PR #573, opened 2014-07-28
Remove dead code from AI_Animal.cpp

cblades:
Remove unreachable code.


xycaleth:
The code removed isn't dead code. If the last element of `mPacks` is removed, then the condition will be true.

PR #579, opened 2014-08-11
[SP] Don't play force/inventory select sound effects while we're in a ci...

eezstreet:
...nematic.


xycaleth:
Closing old pull request.

But it does pose a question: should the player even be able to scroll through their inventory/weapons when a cinematic is playing? I don't see any reason why they should be able to, and they have no way of knowing which weapon or item they're selecting anyway.

PR #603, opened 2015-01-02
[SP, MP] Replace GPL-3+ sys/sys_loadlib.h with GPL-2+ version from ioquake3

smcv:
The one in OpenJK is functionally equivalent, but has GPL-3+ headers
because it came via one of the Return to Castle Wolfenstein family
of games. This is the only file I found in OpenJK that is not GPL-2
compatible, so it seems sensible to simplify the licensing.

(Context: I'm looking into packaging OpenJK for Debian contrib, and this is one of the patches in my local queue.)


ensiform:
Err? Its the same file.  Just remove the license header.

ensiform:
Also note, Raven only released JK2 and JKA under v2, not v2+

smcv:
> Err? Its the same file. Just remove the license header.

I went for the "dot all i's, cross all t's" approach (copying files verbatim from ioquake3) to make sure it was 100% obvious that what I was doing was legally OK. If you're happy with this resolution, great, works for me.

> Also note, Raven only released JK2 and JKA under v2, not v2+

Oh? All the files that have licensing boilerplate say

```
This file is part of Jedi Academy.

    Jedi Academy is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 2 of the License, or
    (at your option) any later version.
```

If it's v2-only like the Linux kernel, then that should be something more like

```
    it under the terms of the GNU General Public License, version 2,
    as published by the Free Software Foundation.
```

I can do a branch if necessary.

(Please don't just delete the license-grant boilerplate altogether; I'd like to be able to get OpenJK into Debian contrib eventually, and the archive administrators are probably going to say no if there's no explicit license grant, so it's good that it's there.)

ensiform:
The version in the SP files boilerplate is incorrect because someone added that later on without knowing.  And nobody has bothered to change it.  We're fine with it being nothing there for now as in general there are no license boilerplates in the MP code at all.  The SP port of that specific file was taken prior to the removal of the invalid boilerplate from MP code and had gone unnoticed until now.

So, we are fine with it just being implied and not there anymore for the time being.

PR #604, opened 2015-01-02
Allow using an external libjpeg-turbo or libjpeg8

smcv:
The JPEG code in OpenJK is basically the same as in ioquake3, which requires either its own  bundled copy of IJG libjpeg 8c, an external copy of IJG libjpeg >= 8, or any vaguely recent version of jpeg-turbo with the jpeg8-compatible MEM_SRCDST_SUPPORTED feature enabled.

Distributions like Debian don't like to use embedded/bundled code copies: it makes it harder to fix security bugs. This branch allows the jpeg-8c directory to be unused, or deleted entirely, when the system libjpeg is suitable.

I've made it default to use the bundled libjpeg on Windows or Mac, or a system libjpeg on Linux, since libjpeg is ubiquitous, and most Linux distributions use libjpeg-turbo these days. If you don't like that default, changing it (e.g. to default to the bundled copy everywhere) would be fine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,12 @@ else()
 	option(UseInternalPNG "Whether to use the included libpng instead of a locally installed one" OFF)
 endif()
 
+if(WIN32 OR APPLE)
+	option(UseInternalJPEG "Whether to use the included libjpeg instead of a locally installed one" ON)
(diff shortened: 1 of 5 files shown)


smcv:
This commit is also now part of #605.

PR #605, opened 2015-01-02
Use system zlib headers if we're using system zlib

smcv:
As I mentioned on #604, distributions like Debian don't like using embedded/bundled code copies, so I'm planning to link OpenJK in Debian contrib to the system zlib. It seems better to use headers that correspond exactly to the system zlib, rather than the header that is bundled in OpenJK.

(This also means I can delete the zlib directory entirely in the Debian package, to know for sure that I'm not using the bundled copy and reduce the number of things I have to do a copyright/license audit on...)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,12 @@ else()
 	option(UseInternalPNG "Whether to use the included libpng instead of a locally installed one" OFF)
 endif()
 
+if(WIN32 OR APPLE)
+	option(UseInternalJPEG "Whether to use the included libjpeg instead of a locally installed one" ON)
+else()
+	option(UseInternalJPEG "Whether to use the included libjpeg instead of a locally installed one" OFF)
+endif()
+
 if(APPLE)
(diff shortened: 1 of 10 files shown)


smcv:
This has some trivial conflicts with #604, so if you want to merge both this and #604, please either resolve the conflict manually, or pull smcv/OpenJK@8afd12f (my `system-libjpeg-and-zlib` branch) instead.

ensiform:
Would rather not have to do any sort of manual merging from prs to accept it.

smcv:
If you intend to merge #604, I'm happy to replace this PR with the appropriate branch; or if you don't intend to merge #604 for whatever reason, this one is mergeable. Sorry, I can't provide branches for both possibilities without making assumptions about what you're going to do next :-)

ensiform:
Both are fine.

smcv:
OK, I've replaced the branch for this PR with one that contains both. I'll close #604 as a duplicate or something.

PR #606, opened 2015-01-03
Correct license boilerplate: the majority is GPL-2 not GPL-2+

smcv:
I know this is a very tedious patch and you'd rather be doing something more concretely useful, but this stuff does matter.

I have not changed the files that were already identified as GPL-2+ in the initial codedrop (code\*/mp3code), the files that have been copied in from ioquake3 (code\*/macosx/macosx_glimp.h, code\*/macosx/macosx_qgl.h, code\*/sdl/sdl_snd.cpp, codemp/client/cl_avi.cpp, codemp/qcommon/md4.cpp) or the files under more permissive licenses (zlib, minizip, md5, etc.) because as far as I can see, Raven/Activision are not copyright holders on those files, so their choice of GPL-2 rather than GPL-2+ is not relevant to those files' licenses.

The effective license of the binaries is the strictest license among the sources that went into them, i.e. GPL-2 (only).

### DIFF
--- a/code/Ragl/graph_region.h
+++ b/code/Ragl/graph_region.h
@@ -2,9 +2,8 @@
 This file is part of Jedi Academy.
 
     Jedi Academy is free software: you can redistribute it and/or modify
-    it under the terms of the GNU General Public License as published by
-    the Free Software Foundation, either version 2 of the License, or
-    (at your option) any later version.
+    it under the terms of the GNU General Public License version 2
+    as published by the Free Software Foundation.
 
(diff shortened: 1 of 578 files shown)


smcv:
Reference for these being GPL-2:

http://sourceforge.net/u/richwhitehouse/jediacademy/ci/c141658a69a44c9f986d225a8917b5c84bf7c3d0/tree/

> Activision and Raven are releasing this code for people to learn from
> and play with. This code is copyright Activision 2003. This source is
> released under GNU GPLv2.

PR #623, opened 2015-02-08
avoid NORETURN macro on function pointer

smcv:
__attribute__((noreturn)), the gcc implementation of NORETURN, is fine
on function pointers, but __declspec(noreturn), the MSVC implementation,
is not.

---

See https://github.com/JACoders/OpenJK/commit/4a09de600ce0c1cdc0e03fc134e192077589957c for discussion

### DIFF
--- a/code/game/g_local.h
+++ b/code/game/g_local.h
@@ -452,7 +452,7 @@ void G_SayTo( gentity_t *ent, gentity_t *other, int mode, int color, const char 
 // g_main.c
 //
 void G_RunThink (gentity_t *ent);
-void QDECL NORETURN G_Error( const char *fmt, ... );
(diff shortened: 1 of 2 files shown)


xycaleth:
Thanks for the new patch. I tried your changes locally on my Windows PC this time, and it still fails to compile. In `code/game/g_local.h`, line 455, `NORETURN` needs to come before `QDECL`. Is this a problem for Linux?

smcv:
I think any order works for gcc. `NORETURN void QDECL G_Error( const char *fmt, ... );` or `void QDECL G_Error( const char *fmt, ... ) NORETURN;` is also OK if bd414aa doesn't work for you.

xycaleth:
Looks good now! Thanks again :)

PR #634, opened 2015-03-05
Shared: fixed terminal mode and signal handlers on unix

ouned:
without this atty mode is always disabled on unix and command history
doesn't work

### DIFF
--- a/code/qcommon/common.cpp
+++ b/code/qcommon/common.cpp
@@ -1565,7 +1565,7 @@ Field_Clear
 ==================
 */
 void Field_Clear( field_t *edit ) {
-	edit->buffer[0] = 0;
+	memset(edit->buffer, 0, MAX_EDIT_LINE);
 	edit->cursor = 0;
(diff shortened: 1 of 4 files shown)


xycaleth:
Looks good :) Thanks for the fix @ouned!

ensiform:
CON_Init was before platform init on purpose.

Also, there is a proper signal handler function already.  Just needs to be hooked up in main().

As for the atty check, I'm not even sure it's required because the console handling works fine in ioq3 without it or did we miss something else in port of sys code?

ouned:
https://github.com/ioquake/ioq3/blob/master/code/sys/sys_unix.c#L809

Didn't work for me without Con_Init after PlatformInit because Con_Init uses stdinIsATTY
ioq3 has PlatformInit before Con_Init, too

ensiform:
I know, I moved it before because it technically needs to be initialized really early for the windows conosle to be set up properly for printing.

PR #635, opened 2015-03-07
Add extra Icarus scriptable character facial expressions

Cerezk:
Finally got around to figuring Git out, and uploading these changes. They are a couple of extra lines of code added to the game that utilise the original game animations to create a wider range of character facial expressions for cinematic cutscenes in the game, scriptable through Icarus. 

Just like the original SET_FACEFROWN, these facial expressions are available through the following commands: 
SET_FACESMILE -- smile 
SET_FACEGLAD -- smile with eyes closed, trusting 
SET_FACEHAPPY -- open smile, happy face 
SET_FACESHOCKED -- mouth open in shock/surprise

### DIFF
--- a/code/game/Q3_Interface.cpp
+++ b/code/game/Q3_Interface.cpp
@@ -326,6 +326,10 @@ stringID_table_t setTable[] =
 	ENUM2STRING(SET_FACEBLINK),
 	ENUM2STRING(SET_FACEBLINKFROWN),
 	ENUM2STRING(SET_FACEFROWN),
+	ENUM2STRING(SET_FACESMILE),
+	ENUM2STRING(SET_FACEGLAD),
+	ENUM2STRING(SET_FACEHAPPY),
+	ENUM2STRING(SET_FACESHOCKED),
 	ENUM2STRING(SET_FACENORMAL),
 	ENUM2STRING(SET_FACEEYESCLOSED),
 	ENUM2STRING(SET_FACEEYESOPENED),
@@ -6031,6 +6035,28 @@ static void Q3_Face( int entID,int expression, float holdtime)
 		ent->client->facial_anim = FACE_FROWN;
 		break;
 
+	//Extra facial expressions:
+	case SET_FACESMILE:
+		ent->client->facial_blink = -(level.time + holdtime);
+		ent->client->facial_timer = -(level.time + holdtime);
+		ent->client->facial_anim = FACE_SMILE;
+		break;
+	case SET_FACEGLAD:
+		ent->client->facial_blink = 1;
+		ent->client->facial_timer = -(level.time + holdtime);
+		ent->client->facial_anim = FACE_TALK1;
+		break;
+	case SET_FACEHAPPY:
+		ent->client->facial_blink = -(level.time + holdtime);
+		ent->client->facial_timer = -(level.time + holdtime);
+		ent->client->facial_anim = FACE_TALK1;
+		break;
+	case SET_FACESHOCKED:
+		ent->client->facial_blink = -1;
+		ent->client->facial_timer = -(level.time + holdtime);
+		ent->client->facial_anim = FACE_TALK3;
+		break;
+
 	case SET_FACENORMAL:
 		ent->client->facial_timer = level.time + Q_flrand(6000.0, 10000.0);
 		ent->client->facial_blink = level.time + Q_flrand(3000.0, 5000.0);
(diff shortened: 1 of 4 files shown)


xycaleth:
On the MP side, you've made changes to the engine which require changes in the mod code. Unfortunately, this will break compatibility with existing mods as they don't use your modified icarus header.

I'm not sure there is any backwards-compatible way of adding these changes. Even if you added the `SET_` enums at the end of the list, existing mods won't be happy with new icarus scripts which make use of these enums (it won't recognise them).

Cerezk:
I haven't experienced any backwards compatibility issues with the base JKA game, Xycaleth. It may be possible that the modifications to the MP code are not needed at all. 

When compiled, all that is needed for the new facial animations to work is the jagamex86.dll, so the changes really only affect the single-player game, and should be backwards compatible with all existing SP mods. Mods whose scripts utilise the new commands/tags will display the extra animations in the new version of OpenJK, while in the older versions, the characters' faces will simply display as much animation as the older version allows. 

It may be that I have submitted a little more code changes than I needed to...

xycaleth:
Your changes for SP are fine, and I don't think will cause any issues because we've broken compatibility with the retail jagamex86.dll already.

The problem is on the MP (multiplayer) side. Anything in the `codemp/game/` folder is compiled into the mod's game DLL (jampgamex86.dll).

Cerezk:
In that case we should leave out the codemp changes, since in all my tests jampgamex86.dll was not needed at all to make it work.

I have only added these changes now, following the pattern of the original code, but I think we only really need the changes in "code" (SP). There are no cinematics in MP, so my additions there were redundant, really.

xycaleth:
Sounds good to me. If you make another change to remove your changes in the `codemp/` folder, it should be good to merge. I'll let @ensiform and @Razish have a quick look over as well in case I missed anything.

Cerezk:
Sure. Sounds good. I'll have to resubmit the whole thing, right? We can't just edit it? 

This will be a great addition for any SP modders out there looking to create a more cinematic experience in their mods. And, since it is just a few extra lines of code to jagamex86.dll, and no edits to any of the existing code, it should be a 100% backwards compatible with everything older. It's just a little extra functionality. An update, if you will.

But feel free to test it as much as possible before committing to it. I'll help here I can.

Cerezk:
Figured it out. Removed the MP code additions.

Cerezk:
Just tested the code in action with the codemp changes removed, and yes it works just fine. The codemp changes were not needed, only code/game/Q3_Interface.cpp and its header, code/game/Q3_Interface.h, need the added lines. 

Changes ready for merge when approved.

ensiform:
Could this still affect saves in any way?

Cerezk:
Not to my knowledge. And I've had this code change running on my system for a year now. Why would it affect saves? It only defines additional Icarus commands.

Not sure what you mean, @ensifrom. There are no changes to the MP code in this commit anymore. I've removed them.

Cerezk:
What we'll need to do is inform people of the additional Icarus commands that can be used, since the SDK documentation only mentions SET_FACEFROWN -- as that's the only facial expression characters could have before this update.

(There is also a random expression script that runs on their face, but that's hardly relevant to making characters express specific emotions.)

ensiform:
The history of your commits is still there.

Cerezk:
Okay... so, what do I do? Sorry, @ensiform, I'm not that familiar with how Git works. The files are MP files are not affected anymore. At least they are not in the "to be changed" list, so I guess my extra two commits (which are to undo the MP code changes) have no effect, then. Why is the history a problem? I don't really understand...

xycaleth:
Personallly I don't think it's a problem. The version history should show all the changes you made, whether they're ugly or not :P

Regarding save file compatibility, the save files are very prone to not being loadable when structs are changed (different padding/alignment/sizes etc between compilers). I might wait until this weekend to merge in your changes, after I've created a working branch for the first release. Any problems with your changes will then have time to be caught before we do a second release :)

Cerezk:
Okay. No problem. That sounds like a wise step. Although, like I said, this small update to the code has been well tested over the course of the last year, and it certainly hasn't posed any threats to stability -- neither with the retail release of the game, nor in OpenJK. The code additions are really quite minimal and straightforward: just an expansion of the existing set of Icarus commands, following the same markup/pattern/logic as the original set. But, having more experience with the code, I'll leave it up to you to do what you think is best. 

I'm just glad I got around to submitting it back to the source code, finally. I'm sure it will serve well for many and help to improve on the wooden performances in JKA cinematics at least a little for all future SP story mods, and maybe even the main game's story -- if someone gets around to re-scripting the original dialogues. 

I am also hopeful that new expressions/animations can be discovered in the future, and this initial code will be expanded, to leap JKA into a more advanced state of animation. That's why I've decided to contribute back to the source. I know OpenJK is staying clear of any serious modifications, but a little update to the old code doesn't hurt to improve the playing experience, especially if it's just an extension of the existing set of rules and features OpenJK has already had in its vanilla code, like this one. 

But I'm rambling...

Issue #628, opened 2015-02-10
Source code licences


xycaleth:
We need to smarten up (read: add) the headers on all of the source files to show that they're licenced under GPLv2. I'm not sure who to list as the original copyright holder though as it's pretty unclear who holds the copyright over the source code :/

Razish:
It's also common to not tag source files, and just have the LICENCE in the root directory for each module.

ensiform:
We already have the license headers in the SP files so it only makes sense for MP as well.

Razish:
Aren't the SP ones wrong?

ensiform:
No not really anymore. @smvc fixed them

smcv:
> It's also common to not tag source files, and just have the LICENCE in the root directory for each module.

That's commonly done, but is far from ideal. Putting a license in the same directory as some code is a lot less clear than specifically saying that (some of? all of?) the code is under the license; per-file license boilerplate makes it unambiguous.

It would also be good to have a verbatim quote of whatever license info was in the original Raven code-drop - I think http://sourceforge.net/u/richwhitehouse/jediacademy/ci/master/tree/README.txt might be it? - somewhere in the documentation, to have as much evidence as possible that we have valid permission from the copyright holders to distribute under GPL-2.

(I'm partly saying this because it's general best-practice, and partly because at some point I want to get OpenJK into Debian contrib, which means convincing the archive administrators that it isn't an unacceptable legal risk.)

smcv:
> I'm not sure who to list as the original copyright holder though as it's pretty unclear who holds the copyright over the source code :/

If in doubt, it's safest to assume that both Activision and Raven are copyright holders.

Here's my best guess at the complete copyright/licensing information (please let me know if you are aware of any other copyright holders or applicable licenses):
http://anonscm.debian.org/cgit/pkg-games/openjk.git/tree/debian/copyright

ensiform:
I added raven's readme in c5940062f61c6948af335fbb426e61c36e2109d8 as rv-readme.txt.

xycaleth:
Following @smcv's lead on this one, the copyright holders are...

id Software
Raven Software
Activision

Is it correct to list OpenJK as a copyright holder? OpenJK isn't a legal entity as far as I understand.

xycaleth:
To clarify, the majority of the files will have the above companies as the copyright holders. We'll still need to go through all of the files to check which ones are relevant.

smcv:
[...]

PR #641, opened 2015-03-18
Change Raven and Activision copyright to include older years

smcv:
Some of the older copyright boilerplate removed by 59dfa47 said
things like

// Copyright 2001-2013 Raven Software
// Copyright 2000-2001 Raven Software
// Copyright 2001-2013 Activision

so we should preserve that. It seems safest to assume that this
applies to to everything.

Change automated by:

find * -type f -print0 | xargs -0 perl -p -i -e \
's/(Copyright \(C\) )2002( - 2013, Raven Software, Inc\.)/$1."2000".$2/eg'
find * -type f -print0 | xargs -0 perl -p -i -e \
's/(Copyright \(C\) )2002( - 2013, Activision, Inc\.)/$1."2001".$2/eg'

Refs #628

### DIFF
--- a/code/Ragl/graph_region.h
+++ b/code/Ragl/graph_region.h
@@ -1,7 +1,7 @@
 /*
 ===========================================================================
-Copyright (C) 2002 - 2013, Raven Software, Inc.
-Copyright (C) 2002 - 2013, Activision, Inc.
+Copyright (C) 2000 - 2013, Raven Software, Inc.
+Copyright (C) 2001 - 2013, Activision, Inc.
 Copyright (C) 2013 - 2015, OpenJK contributors
 
 This file is part of the OpenJK source code.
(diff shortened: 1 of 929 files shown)


xycaleth:
I would say the copyright years for Activision and Raven should be the same as they're effectively the same company. Are we bothered that much?

smcv:
There were no files that said "(c) 2000-20xx Activision" and some files that said "(c) 2000-20xx Raven". I don't know what the exact copyright ownership situation is: if Raven is a subsidiary of Activision, that doesn't *necessarily* imply that they can't be a separate copyright holder, AIUI.

But I can easily run it through another sed command if that's what you want.

xycaleth:
I guess we can just leave it as it is then :)

PR #642, opened 2015-03-18
Add support for using a system copy of minizip

smcv:
Linux distributions prefer to use a shared system copy of each library
rather than having projects bundle it, so that any security fixes
or other bugfixes only need to happen once. minizip is in at least
Debian, Ubuntu and Fedora (they need it for Chromium).

I'm using pkg-config to find it, on the assumption that the bundled
internal copy will continue to be used on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 2.8.8)
 # For checks in subdirectories
 set(InOpenJK TRUE)
 
+include(FindPkgConfig)
+
 # Project name
 set(ProjectName "OpenJK" CACHE STRING "Project Name")
 project(${ProjectName})
@@ -30,6 +32,8 @@ if(WIN32)
 	option(UseInternalOpenAL "Whether to use the included OpenAL instead of a locally installed one" ON)
 	option(UseInternalZlib "Whether to use the included zlib instead of a locally installed one" ON)
 endif()
+
+option(UseInternalMinizip "Whether to use the included minizip instead of a locally installed one" ON)
 
 if(WIN32)
 	option(UseInternalPNG "Whether to use the included libpng instead of a locally installed one" ON)
@@ -260,6 +264,10 @@ endif()
 
 if(UseInternalZlib)
 	set(SharedDefines ${SharedDefines} "USE_INTERNAL_ZLIB")
+endif()
+
+if(UseInternalMinizip)
+	set(SharedDefines ${SharedDefines} "USE_INTERNAL_MINIZIP")
 endif()
 
 set(OpenJKLibDir "${CMAKE_SOURCE_DIR}/lib")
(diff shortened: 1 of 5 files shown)


ensiform:
I'm not sure I want to allow this as it removes the usage of Z_Malloc/Z_Free in the minizip files.

smcv:
> removes the usage of Z_Malloc/Z_Free in the minizip files

Hmm, fair point.

I'll have to dig into exactly how those are used: if it's a short-term allocation (malloc, use, free) then IMO it's fine to use a system allocator, but if it leaves the scope of that file it's not OK.

smcv:
Expanding on that a little, am I right in saying that Z_Malloc/Z_Free is basically a "memory pool" implementation, so that the memory allocated for a level is bounded, and can all be discarded when leaving the level (by error or otherwise) without having to free individual allocations?

I don't think bounded memory is as useful now as it was when Mac OS 9 was a thing; modern operating systems don't impose an up-front memory limit.

The memory-pool part is obviously still useful, but for short-term allocations "behind the scenes", a malloc/use/free/return isn't really any better or worse than stack-based allocations than alloca/use/return.

ensiform:
They point to Malloc somewhat but it uses a tag system with blocks to track the memory usage.

ensiform:
It isn't the same as it is in q3.

smcv:
Looks like OpenJK's minizip does return Z_Malloc'd memory, so, you're right, it's a slightly incompatible fork rather than just a minizip subset, at least unless/until minizip gets a "custom malloc" hook like [the one in GLib](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-mem-set-vtable).

With that in mind, it might make sense to move it from lib/ to shared/ and add a README pointing out that it isn't a drop-in replacement?

ensiform:
I really see no point in moving it.  It's still a lib after all.  Other tham making a mention in the file or readme to state why no system lib.

Also, I have doubts that newer versions (1.1+) of mini(un)zip would be compatible due to the changes to support larger zips and 64 bit integers plausibly breaking checksum matching.

xycaleth:
I agree with @Ensiform that minizip should stay in the `lib/` folder. I do think it's a good idea to point out that modifications were made to it (and the same should be done to the other libraries as well).

smcv:
The others don't seem to have incompatible modifications:

jpeg-8c is the same as IJG libjpeg 8c, except for trivial compiler-warning fixes (s/main/mainptr/ mainly).

zlib is zlib-1.2.7 with nonessential files deleted.

libpng is libpng-1.6.16 with nonessential files deleted and `pnglibconf.h` (normally a generated file) added.

SDL2 has some diffs in `include/SDL_config.h` but, again, that's a generated file.

xycaleth:
For MiniZip, it seems you can define `ALLOC` to use your own allocator, but it's not obvious how to define this so that the source file can make use of it...

smcv:
> it seems you can define ALLOC to use your own allocator

That's compile-time configuration, so it doesn't help us to use a system copy of minizip (which will have been compiled with ALLOC = malloc, for use in Chromium etc.). It would only be useful for the Q3 engine if it was runtime configuration, like in GLib.

smcv:
ioquake3 upstream have now merged this commit from me:

https://github.com/ioquake/ioq3/commit/caf08fdb7b939e72bbb49a15e06c4ce00b214a3c

to clarify that a system minizip is not supported, and perhaps more importantly, why not. I'll prepare an equivalent patch for OpenJK at some point if nobody gets there first.

Issue #643, opened 2015-03-19
Linux cgame, ui, jampgame DLLs no longer installed by default


smcv:
Last time I did a build for Debian (from commit 471dd02) I could configure cmake with

		-DBuildJK2SPEngine=ON \
		-DBuildJK2SPRdVanilla=ON \
		-DBuildJK2SPGame=ON \
		-DCMAKE_BUILD_TYPE=RELWITHDEBINFO \
		-DCMAKE_INSTALL_PREFIX=/usr/lib/openjk \
		-DCMAKE_VERBOSE_MAKEFILE=TRUE \

and then `make install DESTDIR=$(pwd)/debian/tmp` would leave me with these files in debian/tmp, where `*` is a wildcard for architecture-dependent bits and `.so` is the Linux equivalent of `.dll`:

```
usr/lib/openjk/openjk_sp.*
usr/lib/openjk/rdsp-*.so
usr/lib/openjk/OpenJK/jagame*.so
usr/lib/openjk/openjk.*
usr/lib/openjk/rd-*.so
usr/lib/openjk/openjkded.*
usr/lib/openjk/OpenJK/cgame*.so 
usr/lib/openjk/OpenJK/ui*.so
usr/lib/openjk/OpenJK/jampgame*.so
usr/lib/openjk/openjo_sp.*
usr/lib/openjk/rdjosp-vanilla_*.so
usr/lib/openjk/OpenJK/jospgame*.so
```

(and then I split up JA and JK2 between base directories /usr/lib/openjk-academy and /usr/lib/openjk-outcast so they don't try to use each other's files, but that's in the Debian packaging not the upstream bit).

In current builds I get a new empty zip file at `openjk.pk3`, but I no longer get the cgame, ui or jampgame DLLs.

Unpacking DLLs from the pk3 file is not implemented on Linux, so it's OK that it's empty - as far as I'm concerned, if this worked, it would be a security vulnerability (because auto-downloading exists, and arbitrary native code execution from unauthenticated auto-downloaded packages is a bad thing) so I would like to continue to use DLLs that are ordinary files in the filesystem.

smcv:
I'll try to look into this this evening if nobody has spotted the cause by then - all three MP-relevant DLLs are missing, so it looks as though maybe `make install` isn't recursing into `codemp` correctly?

smcv:
[...]

PR #644, opened 2015-03-19
[MP] Install game, cgame, ui DLLs on non-Mac again

smcv:
See #643.

This puts it back the way it was before d4cc979: alternatively we could omit the DLLs on Windows, and/or install to both `base/` and `OpenJK/` (for some reason before d4cc979 this was still done for Mac bundles, but not Windows or Linux).


smcv:
Pushed an updated version which uses WIN64.

xycaleth:
Looks good!

PR #645, opened 2015-03-20
please add licenses of cmake modules

smcv:
Some of the cmake modules point to third-party licenses which aren't included in the source tree. It's better to include all applicable licenses with the source, so that the conditions for redistribution for the whole project are self-contained.

In this pull request I've followed the instruction in the cmake-licensed modules to "substitute the full License text for the above reference". If you'd prefer to include Copyright.txt as a separate file instead, you could merge my cmake-licenses-alt branch instead: https://github.com/JACoders/OpenJK/compare/JACoders:master...smcv:cmake-licenses-alt


xycaleth:
Thanks again for your work :) About the license in InstallZIP.cmake, I wasn't sure which one to use there so I left it as it is. The file itself is based off of a CMake module, but I've edited it a fair amount. This is the original file: https://github.com/Kitware/CMake/blob/master/Modules/CPackZIP.cmake

smcv:
It would be good to mention "based on CPackZIP.cmake from CMake x.y.z" in InstallZIP.cmake, if that's the case.

> Is it fine to continue using that license?

The copyright holders of the new file are the copyright holders of the file it was based on (in this case Kitware, Inc.), and you; so you can do anything for which both you and Kitware give permission. The license grant from Kitware says you can modify and redistribute if you follow the conditions in the license they gave, which mostly means including a copy of that license.

The license grant from you ... well, what do you want to allow? :-)

I think the most sensible option is to add yourself as an additional copyright holder in that file (under your real name, or a pseudonym, or as "OpenJK contributors"), and keep using the CMake license for it.

If you strongly prefer the GPL, you could say something like

```
# This file is part of Jedi Academy.
# Jedi Academy is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License version 2
# as published by the Free Software Foundation.
# Jedi Academy is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
# GNU General Public License for more details.
# You should have received a copy of the GNU General Public License
# along with Jedi Academy. If not, see <http://www.gnu.org/licenses/>.
# 
# Based on CPackZIP.cmake which is under the following license:
# 
# Redistribution and use in source and binary forms, with or without
# modification, are permitted provided that the following conditions
# are met:
...
```

but that adds yet more boilerplate and tbh I think that would just be making life more difficult for yourself.

xycaleth:
Hopefully I'll get all of this copyright stuff eventually :) It's one thing to learn it at university, but it's another thing to actually make use of it in practice! I'm happy for you to just add "OpenJK contributors" as an additional copyright holder. Adding a second license is just more confusing!

smcv:
> I'm happy for you to just add "OpenJK contributors" as an additional copyright holder

Branch updated.

Issue #630, opened 2015-02-13
different icons for JK2/SP, JA/SP, JA/MP?


smcv:
(tl;dr: if someone can give me SVG or layered XCF source for the OpenJK icon, I'll try fixing this myself.)

If you have all the flavours of OpenJK installed, they aren't distinguishable at a glance because there's only one icon, particularly with GNOME Shell's bad habit of truncating app names:

![](http://www.pseudorandom.co.uk/~smcv/20150213openjk/jediicons.c.png)

The original games solved this by having a different icon with multiple lightsabers for MP mode.

Return to Castle Wolfenstein, which has the same "separate engine for SP/MP" issue, has traditionally got around this by having different coloured icons, and iortcw (analogous to ioquake3/OpenJK) does the same for its redrawn icon. For the Debian packaging, I also made it more obvious which colour was multiplayer by editing that icon to have smaller, alpha-blended copies:

![](http://www.pseudorandom.co.uk/~smcv/20150213openjk/wolficons.c.png)

Does someone have a source file for the OpenJK icon, like a SVG or layered GIMP/Photoshop file? I'd like to try doing a similar modification for that. Perhaps JK2/SP could keep the blue-on-black (or get more blue) to be reminiscent of its box art, JA/SP could have a more teal/green version to be reminiscent of *its* box art, and JA/MP could have multiple lightsabers shown?

ensiform:
There's a Photoshop file but it doesn't render in gimp correctly I believe.

ensiform:
I would prefer the main icon remain as is for JAMP.

And Jedi Outcast is not being distributed in the first release so, that is not relevant at this time.  JASP exe could be different color though.

smcv:
Yeah, I'm aware that JO support in OpenJK is experimental. I'll have to include JO in the initial upload to Debian anyway, because adding extra binary packages at a later date is painful (it goes back through legal/technical review every time I add one); but I'll initially only be uploading to the 'experimental' part of the archive, and when the JA part is ready to leave 'experimental', I'll make sure to exclude JO from those builds.

ensiform:
But they are separate games to be run from separate folders.

smcv:
In the current packaging, openjk + openjk_sp are one .deb (openjk-academy), openjkded is another (openjk-academy-server) and openjo_sp is a third (openjk-outcast). openjk-academy and -server share a base directory if they're both installed, but openjk-outcast has a separate base directory.

[...]

PR #647, opened 2015-03-23
Add the OpenJK icon's source material, released under the MIT/X11 license

smcv:
With permission from the artist, Antonio Aiello <http://tristamus.com>.

Related to JACoders/OpenJK#628, JACoders/OpenJK#630.


smcv:
If you don't like `shared/icon` feel free to move it somewhere else... but I wanted to get a PR with the licensing info ready as soon as I could.

xycaleth:
I feel like we need a separate folder for the icons to make it clear that the folder doesn't have any source code. Something like, `/resources/icons/` perhaps?

smcv:
Sure, whatever you like. But if you wouldn't mind merging this first and moving things around later, I think all the pieces are in place for me to take a snapshot of the git repo and do the first fully DFSG-clean Debian packages for this stuff :-)

smcv:
`/resources/icon/`, `/data/icon/`, `/shared/icon/`, or any of those with `icons` instead of `icon` all seem perfectly reasonable places to put this. Any is fine.

I would suggest not following ioquake3's example - `/misc/` is not the most self-describing directory name! I would also suggest avoiding `/assets/` so people don't think "wait, isn't `assets0.pk3` non-distributable?" and worry that it contains some of that stuff.

Potential future work (which I might do one day): make the build system use these icons instead of the ones in OS-specific directories, possibly moving `OpenJK.icns` into the same place as the `.psd` source. But for now the source files (and their resized versions) are just shipped as source and not hooked up to the build system at all.

xycaleth:
Sounds reasonable!

PR #655, opened 2015-04-19
Shared: fixed binding keys a-z in the menu

ouned:
menu code expects uppercase keycodes

### DIFF
--- a/shared/sdl/sdl_input.cpp
+++ b/shared/sdl/sdl_input.cpp
@@ -262,7 +262,11 @@ static fakeAscii_t IN_TranslateSDLToJKKey( SDL_Keysym *keysym, qboolean down ) {
 	if( keysym->sym >= SDLK_SPACE && keysym->sym < SDLK_DELETE )
 	{
 		// These happen to match the ASCII chars
-		key = (fakeAscii_t)keysym->sym;
+		if (keysym->sym >= SDLK_a && keysym->sym <= SDLK_z) {
+			key = (fakeAscii_t)('A' + (keysym->sym - SDLK_a));
+		} else {
+			key = (fakeAscii_t)keysym->sym;
+		}
 	}
 	else
 	{


xycaleth:
Please can you provide a more detailed description of the bug you're fixing?

ouned:
1) go into openjk
2) try binding chat to the 'u' key via menu not console
3) doesn't work

ensiform:
I thought it already converts down the chain in the client code. Hmm..

ouned:
Since it affects windows, too you need to fix it like this to support legacy mods
It seems to only affect keys which had a previous binding like trying to bind chat to 'w'

xycaleth:
I took a bit of time to investigate this problem a bit further. The problem comes about when binding a key which was already bound to another command. Looking in `Item_Bind_HandleKey`, it has the following code:

```c++
	if ( key != -1 ) {
		size_t	i;

		for ( i=0; i<g_bindCount; i++ )
		{
			if ( g_bindKeys[i][1] == key )
				g_bindKeys[i][1] = -1;

			if ( g_bindKeys[i][0] == key ) {
				g_bindKeys[i][0] = g_bindKeys[i][1];
				g_bindKeys[i][1] = -1;
			}
		}
	}
```
This loop is intended to remove the key from all commands, but only if the key *and case* matches. Later on in `Controls_SetConfig`, if the command that was just bound uses the same key as another command but with a different case, then the second command will overwrite the first.

As a concrete example, the U key is bound to `messagemode3` by default. When attempting to bind the U key to another command such as `+left`, the *lower case* U is associated with that command. The U key is not removed from the `messagemode3` key because it is associated with the `upper case` U. As `+left` appears before `messagemode3` in the list of commands (`g_keyBindings`), `messagemode3` overwrites the bind for the U key and it appears as if the key was never bound.

xycaleth:
So, I forgot to review the code. @ouned  The fix looks fine for characters in the a-z range, but probably still leaves the extended ASCII keys broken. I assigned them all to `A_LOWER_xxx`. Can you edit your fix to change these to `A_UPPER_xxx` please?

ouned:
I don't think this would be enough to get them working.
https://wiki.libsdl.org/CategoryKeyboard
https://wiki.libsdl.org/SDL_Keycode
With my german keyboard on pressing 'ä' it generates this:
scancode = SDL_SCANCODE_APOSTROPHE
keysym->sym = 228

xycaleth:
The key/scancode doesn't matter for key input. It's the `sym` field that matters. 228 is 0xE4 in hexadecimal which is mapped to `A_LOW_ADIEREISS` (which is ä)

PR #658, opened 2015-04-27
fix fallout from removal of ibize_platform.h

### LINKED ISSUE #651: SP won't compile on C++14
Due to changes with string literals in C++14, the compile will fail.

error C3688: invalid literal suffix 'S_COLOR_GREY'; literal operator or literal operator template 'operator ""S_COLOR_GREY' not found	cvar.cpp:1013

Here's a sample of one of the lines:
`Com_Printf( S_COLOR_GREY"Bind "S_COLOR_WHITE"%s = "S_COLOR_GREY"\""S_COLOR_WHITE"%s"S_COLOR_GREY"\""S_COLOR_WHITE"\n", Key_KeynumToString( b ), kg.keys[b].binding );`

smcv:
After commit 03235915 removed the inclusion of ibize_platform.h, I get compiler errors in codeJK2 on Linux (gcc).

One of the commits here should also fix #651.

### DIFF
--- a/code/game/g_savegame.cpp
+++ b/code/game/g_savegame.cpp
@@ -499,7 +499,7 @@ static void EnumerateField(const save_field_t *pField, const byte *pbBase)
 	}
 }
 
-static void EnumerateFields(const save_field_t *pFields, const byte *pbData, unsigned int ulChid, int iLen)
+static void EnumerateFields(const save_field_t *pFields, const byte *pbData, unsigned int ulChid, size_t iLen)
 {
 	strList = new std::list<sstring_t>;
 
--- a/codeJK2/cgame/cg_weapons.cpp
+++ b/codeJK2/cgame/cg_weapons.cpp
@@ -215,11 +215,11 @@ void CG_RegisterWeapon( int weaponNum ) {
 	// give ourselves the functions if we can
 	if (weaponData[weaponNum].func)
 	{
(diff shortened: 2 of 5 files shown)


ensiform:
Why was cstdint specifically added to JK2 ?

smcv:
> Why was cstdint specifically added to JK2 ?

I'm not sure why specifically `cstdint`. The high-level reason was to get `uint32_t` defined so that `ParseRGB()` could return one instead of a (typedef for a) DWORD, when the single #include of `ibize_platform.h` was removed from JK2: nothing else depended on that file, or on anything else from `tools`, so getting rid of that #include was a good thing overall.

The rest of the codebase uses the C `stdint.h` to get the [u]intN_t family of types. That is, strictly speaking, not portable (it isn't in the subset of C that C++ compilers guarantee to provide, but in practice they all do); `cstdint` is the standard way to get `uint32_t` defined by a C++ compiler.

smcv:
`codeJK2: replace __cdecl with QDECL` is a remaining issue related to disentangling JK2 from `ibize_platform.h`.

The other two commits here are simple fixes for compiler warnings (`g_savegame...`) and errors (`icarus...`) which might be related to removal of that header, or not - I'm not sure. I can do a separate pull request per fix if you want, but that's probably more time consuming (for you and for me), so I'd rather continue to group simple fixes together.

I've dropped the #651 part from this pull request, since removing the use of `cstdint` downgrades its severity from "can't compile without fixing this" to "would be nice to fix one day".

PR #659, opened 2015-04-30
Use C++11-compliant string concatenation

smcv:
C++11 reserves syntax like X"foo" and "foo"X for future syntax like
L"a wide-char string", so concatenating macros with literals now
requires an intervening space. Fixes most, if not all, of #651.

---

Tested on Linux, so I could potentially have missed some instances of non-C++11-compliant concatenation in code specific to other platforms - compiling with MSVC in C++11 or C++14 mode would probably detect anything that's left.

### DIFF
--- a/code/client/cl_keys.cpp
+++ b/code/client/cl_keys.cpp
@@ -1026,7 +1026,7 @@ void Key_Bind_f( void ) {
 
 	if ( c == 2 ) {
 		if ( kg.keys[b].binding && kg.keys[b].binding[0] )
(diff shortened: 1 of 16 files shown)


ensiform:
MSVC doesn't have a mode for either 11 or 14.  It just is or it isn't depending on the version. AFAIK.

smcv:
(pushed an updated version that does not include the commits from #658)

smcv:
> MSVC doesn't have a mode for either 11 or 14. It just is or it isn't depending on the version. AFAIK.

Ah, OK, I'd misunderstood #651. In that case, the way to detect whether there's anything left to do is for @eezstreet to compile on whatever the relevant version is :-)

PR #665, opened 2015-05-08
MP: PM_VehicleViewAngles: fix use of uninitialized variables

smcv:
We shouldn't use clampMin or clampMax until they have been set,
at which point setAngles is set to true. Initializing it to true
defeats that.


ensiform:
Certain this won't affect the cgame<>game being in sync?

smcv:
In the cases where it matters (if such cases exist in practice), they'd be out of sync anyway, because the game code would be using uninitialized stack memory for the view angles; that seems unlikely to yield consistent results.

xycaleth:
I think this change makes sense. Without the change, the client and server may have been out of sync anyway if neither `clampMin` nor `clampMax` were set.

Issue #637, opened 2015-03-08
r_gamma affects entire screen


xycaleth:
Setting `r_gamma` affects the entire display as setting the gamma ramp affects the display owning the game window. To improve this, we could apply the gamma correction ourselves in the renderer as a "post processing" step so that it only affects the rendered image, instead of the whole display.

For the vanilla renderer, we would need to check for availability of ARB shaders (should be available on every PC nowadays), and then apply the gamma correction in a pixel shader. Steps to take:

1. Copy back buffer to texture using `glCopyTexImage2D`.
2. Draw texture to screen as full-screen quad, applying pixel shader to gamma correct the texture.

sudge:
I have an application which can fix gamma without system reboot. https://www.dropbox.com/s/6icc005yj0ilria/fixgamma.rar?dl=0 fixgamma.exe in archive. I have no idea about virus check, you can try yourself.

PR #667, opened 2015-05-16
post-processing gamma correction replacement using glsl and framebuffers

ouned:
this is the post-processing gamma correction replacement I made for jk2mv because I was so annoyed by the hardware gamma. It uses extensions and falls back to hardware gamma if not supported. It has problems on linux with drivers using a mesa3d opengl implementation, thus is disabled there by default.

targets issue #637


xycaleth:
I'm going to say I don't want to accept this pull request, sorry :( There's a few issues I have with it:

1. Framebuffers aren't necessary for this. You can copy the back buffer to a texture, and then draw this texture with the gamma correction shader back to the back buffer.
2. The vanilla renderer already makes use of the ``GL_ARB_vertex_program`` and ``GL_ARB_fragment_program`` extensions (for dynamic glow). I would really prefer if the gamma correction shader used these extensions instead of the GLSL ones. The shaders are bit more difficult to write, but it means more people are able to use it.

ouned:
I knew that :p
Im just looking to improve jk2mv, collecting suggestions

I agree with 1. but i have no clue how that would work (i never really did any OpenGL stuff this was my first ogl code)
2. glsl is OpenGL 2.0 which is already out for 10 years or so?

ensiform:
Because we don't want to increase the graphics card requirements for the vanilla renderer.  That's the point of alternate renderers.

Razish:
+1 on Xycaleth's + Ensiform's comments

xycaleth:
> I agree with 1. but i have no clue how that would work (i never really did any OpenGL stuff this was my first ogl code)

You can use the [`glCopyTexImage2D`](https://www.opengl.org/sdk/docs/man3/xhtml/glCopyTexImage2D.xml) function. For example...

```c++
// After you've rendered everything...

glBindTexture(GL_TEXTURE_2D, tr.gammaRenderTarget);
glCopyTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA8, 0, 0, glConfig.vidWidth, glConfig.vidHeight, 0);

// At this point, you now have the screen pixels in the tr.gammaRenderTarget texture

// Draw the texture to the screen again with your GLSL program as you were before
```
No need for the framebuffer at all here.

PR #672, opened 2015-05-19
Update NPC_stats.cpp

Brownjor10:
I've raised the npc memory to 0x80000 and the models per level to 60.  If this creates a problem with slowing the game, we can always go a little lower.  But, I doubt it should, unless limits are actually hit in game.  I'm new to coding, excuse me.  I'm just really hoping we can raise the npc limit for the game a little more, due to a mod I have been working on since before the source code was released.

### DIFF
--- a/code/game/NPC_stats.cpp
+++ b/code/game/NPC_stats.cpp
@@ -40,7 +40,7 @@ extern vec3_t playerMins;
 extern vec3_t playerMaxs;
 extern stringID_table_t WPTable[];
 
-#define		MAX_MODELS_PER_LEVEL	40
+#define		MAX_MODELS_PER_LEVEL	60
 
 hstring		modelsAlreadyDone[MAX_MODELS_PER_LEVEL];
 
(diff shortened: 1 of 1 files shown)


ensiform:
I think the NPC data size is already this large in MP so it should be good to merge

Brownjor10:
Awesome.  Who adds it?  Does it get confirmed by someone or what?

ensiform:
I don't have access to a computer right now, so I can only make changes like closing and stuff or adding comments.

Someone else will need to accept/merge.

xycaleth:
> On 24 May 2015, at 00:14, Ensiform <notifications@github.com> wrote:
> 
> I don't have access to a computer right now, so I can only make changes like closing and stuff or adding comments.
> 
> Someone else will need to accept/merge.
> 
I'll have a look at this tomorrow morning (GMT time).

Brownjor10:
Oh, you guys are good.  I'm just learning how the site works.  Thanks for helping.

xycaleth:
Looks fine to me. We may want to consider in the future redesigning how NPCs are loaded so we don't have to keep them all in a massive string... same goes for shaders etc

PR #674, opened 2015-06-08
Try to load entities from separate file instead of BSP

therealmik:
Since servers are in control of entities, a server can customize base maps by editing the entities chunk of the bsp to add/remote items, spawnpoints, etc.

This patch makes it a bit easier - simply put in your usual path (.pk3 or base/maps/...) a file with .entities tagged onto the end of the name (eg. maps/mp/ffa3.bsp.entities), and you can have your custom entities loaded without changing base assets.

### DIFF
--- a/codemp/qcommon/cm_load.cpp
+++ b/codemp/qcommon/cm_load.cpp
@@ -455,7 +455,30 @@ void CMod_LoadBrushSides (lump_t *l, clipMap_t &cm)
 CMod_LoadEntityString
 =================
 */
-void CMod_LoadEntityString( lump_t *l, clipMap_t &cm ) {
+void CMod_LoadEntityString( lump_t *l, clipMap_t &cm, const char* name ) {
+	fileHandle_t h;
+	const int entNameLen = strlen(name) + 1;
+	char* entName = (char *)malloc(entNameLen);
+	if(entName != NULL) {
+		Q_strncpyz(entName, name, entNameLen);
+		entName[entNameLen-4] = 'e';
+		entName[entNameLen-3] = 'n';
+		entName[entNameLen-2] = 't';
+		const int iEntityFileLen = FS_FOpenFileRead( entName, &h, qfalse );
+		if (h)
+		{
+			cm.entityString = (char *)Hunk_Alloc( iEntityFileLen+1 , h_high );
+			cm.entityString[iEntityFileLen] = '\0';
+			FS_Read( cm.entityString, iEntityFileLen, h);
+			FS_FCloseFile( h );
+			cm.numEntityChars = strlen(cm.entityString);
+			Com_Printf("Loaded entities from %s\n", entName);
+			free(entName);
+			return;
+		}
+		free(entName);
+	}
+	
 	cm.entityString = (char *)Hunk_Alloc( l->filelen, h_high );
 	cm.numEntityChars = l->filelen;
 	Com_Memcpy (cm.entityString, cmod_base + l->fileofs, l->filelen);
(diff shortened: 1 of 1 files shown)


codemp/qcommon/cm_load.cpp
```diff
@@ -455,10 +455,32 @@ void CMod_LoadBrushSides (lump_t *l, clipMap_t &cm)
...
+                Q_strcat(entName, entNameLen, ".entities");
+
+                const int iEntityFileLen = FS_FOpenFileRead( entName, &h, qfalse );
+		Com_Printf("Trying to Load entities from %s\n", entName);
+                if (h)
+                {
+                        cm.entityString = (char *)Hunk_Alloc( iEntityFileLen , h_high );
+                        FS_Read( cm.entityString, iEntityFileLen, h);
+                        FS_FCloseFile( h );
+                        Com_Printf("Loaded entities from %s\n", entName);
```
dionrhys: Should `cm.numEntityChars` not be set to `iEntityFileLen` here?

codemp/qcommon/cm_load.cpp
```diff
@@ -455,10 +455,32 @@ void CMod_LoadBrushSides (lump_t *l, clipMap_t &cm)
 CMod_LoadEntityString
 =================
 */
-void CMod_LoadEntityString( lump_t *l, clipMap_t &cm ) {
-	cm.entityString = (char *)Hunk_Alloc( l->filelen, h_high );
-	cm.numEntityChars = l->filelen;
-	Com_Memcpy (cm.entityString, cmod_base + l->fileofs, l->filelen);
+void CMod_LoadEntityString( lump_t *l, clipMap_t &cm, const char* name ) {
+        fileHandle_t h;
+        const int entNameLen = strlen(name) + 10;
```
xycaleth: ...why `+ 10`?

dionrhys: Because of the concatenation of ".entities" and the nul byte.

shinyquagsire23: Maybe an strlen(".entities") + 1 wouldn't hurt there in that case.

therealmik:
Both of your comments were spot on.  Fixed it and removed a couple of debug prints that wouldn't interest an end-user.

dionrhys:
I maybe really pedantic now, but you could just have `const char suffix[] = ".entities"` and use `sizeof(suffix)` to get the size of the string (with nul byte) at compile time :)

Do others agree on the file extension ".entities" rather than the ".ent" extension commonly used by mods like Lugormod and the Q3Map2 compiler? I'd prefer .ent myself, but then again map developers may already have stray .ent files around in the maps folder from previous decompilation and -onlyents compiles that they don't wish to use to override the entities at run-time.

Razish:
I would recommend having `.ent`, that was my first reaction when I saw the code.

ensiform:
Should this be squashed down too before at we merge it ?

codemp/qcommon/cm_load.cpp
```diff
@@ -455,7 +455,30 @@ void CMod_LoadBrushSides (lump_t *l, clipMap_t &cm)
...
+		Q_strncpyz(entName, name, entNameLen);
+		entName[entNameLen-4] = 'e';
+		entName[entNameLen-3] = 'n';
+		entName[entNameLen-2] = 't';
+		const int iEntityFileLen = FS_FOpenFileRead( entName, &h, qfalse );
+		if (h)
+		{
+			cm.entityString = (char *)Hunk_Alloc( iEntityFileLen+1 , h_high );
+			cm.entityString[iEntityFileLen] = '\0';
+			cm.numEntityChars = strlen(cm.entityString);
```
therealmik: Sorry, this is wrong, will resend. (strlen needs to be after read)

dionrhys:
I've merged it in with some alterations made so it uses static memory rather than malloc :)

PR #679, opened 2015-06-18
Make bundled zlib/minizip static libs.

ahmedtd:
Cleans up the logic around switching between bundled and system
zlib/minizip.

Previously, minizip was always bundled; now it is controlled by the zlib
selection (for a long time now, minizip has been distributed within the
zlib release tarballs, so it doesn't make sense to support a split
setup).

Previously, bundled zlib and minizip were directly added to the projects
that needed them as source files.  This required some modifications to
the minizip code.  Now, both are built as position-independent static
libraries and linked into the target that need them.  This reduces the
cmake logic required for each target, and also lets us use an unmodified
copy of minizip.

As a consequence of this change, the flag "-fvisibility=hidden" is no
longer set project-wide, but as a per-target property where it is
needed.  zlib defaults to visible symbols and manually hides some of
them, rather than the typical method of defaulting to hidden symbols and
manually exporting some of them.

Additionally, the minimum cmake version has been bumped to 2.8.9, since
that is the first version that has the POSITION_INDEPENDENT_CODE target
property, which allows portably specifying that a target should be
created with position-independent code.

This change has been tested on linux x86_64 (gcc).  Care was taken to
not break Windows or clang compatibility, but I don't have Windows or
clang dev environments handy for testing.


smcv:
The bundled zlib is unmodified and can be compiled as a static library or avoided altogether. In Debian I used the system zlib.

However, the bundled minizip is modified, and is functionally different: see https://github.com/ioquake/ioq3/commit/caf08fdb7b939e72bbb49a15e06c4ce00b214a3c and #642.

ahmedtd:
Hmm, I didn't realize that.

It's frustrating, because the difference between the upstream and bundled version is so small --- only swapping around some preprocessor definitions to use Z_Malloc.

I'm fine with always using the bundled copy, then.  I should note, though, that the system copy seemed to work just fine (no stuttering or hiccups from malloc), even on my somewhat anemic laptop.

smcv:
As far as I know, the (only) issue with the system minizip is that it leaks memory: for every zip file that was open at a level transition, the memory remains in use but becomes unreachable. If you have plenty of RAM (or plenty of swap) and don't change levels abusively often, it would probably be difficult to notice.

ahmedtd:
Ah, I see.  I will look and see what might be necessary to reclaim it on level switch.

ahmedtd:
Well, I've only looked at the single-player code, but it doesn't look like anything is leaked at level change.  Each pak3 file is opened once in the search path, but the search path is never changed after game startup.  Additional files that are opened as "unique" get their own dedicated minizip structure (and their own call to unzOpen), but FS_FCloseFile properly closes this dedicated instance.

In the single-player codebase, it doesn't look like any subsystem uses FS_FOpenFile without properly closing it.  Some of the closures are a bit indirect (like in snd_dma.cpp), but they all appear to do the right thing.  I still need to look through the multiplayer codebase.

ensiform:
@xycaleth @Razish what do we want with this pull request?

smcv:
I think changing how zlib is linked is a bit of a red herring if what you actually want (like I do, and as e.g. Debian and Fedora policy mandate) is to use system libraries wherever possible. If you want to use a system copy of zlib, just do that, using the option that already exists; the default is to use a system zlib on non-Windows anyway.

The bundled minizip is not in the same situation: it has a functional change, and is also so little code that replacing it with a system copy is not necessarily a big deal anyway. Yes there have been security issues in minizip (which is the strongest justification for avoiding embedded code copies) in the past, but as far as I can tell they have all been path traversal bugs in the accompanying command-line utility, which OpenJK doesn't have, rather than anything in the library code.

> FS_FCloseFile properly closes this dedicated instance

But does that get called at an abnormal level terminations, for instance?

Part of the point of the Z_Malloc allocator is that it's a "memory pool" which supports this usage pattern (pseudocode):

```
alloc(23)
alloc(42)
alloc(1000)
/* ... later, on error... */
dealloc() /* frees all 1065 bytes at once */
```

smcv:
b9251add06f7 and 1db8c6e62 do seem like worthwhile fixes to cherry-pick, although I haven't reviewed them in detail. As far as I can tell, OpenJK mostly avoids the C++ STL (standard template library) so it might be better to use malloc/free instead of std::vector<char> in the SP code path.

ahmedtd:
In the entire codebase, every call to FS_OpenFile is immediately (or nearly immediately) matched by a corresponding call to FS_CloseFile.

Changing the bundled libs to static libraries is good because it prevents them from being compiled twice.

As for avoiding std::vector -- any project is free to have whatever standards it likes, but what is the point of using C++ if you do not allow yourself to use the features it provides?  In this case, use of std::vector is exactly equivalent to malloc+free.

xycaleth:
Sorry I haven't got around to looking at this earlier. I'm all for compiling more of the libraries as static libs and statically linking, but on Linux it seems to me that dynamically linking is more common. When not using system libraries then using a static lib is a good idea :)

> I'm fine with always using the bundled copy, then. I should note, though, that the system copy seemed to work just fine (no stuttering or hiccups from malloc), even on my somewhat anemic laptop.

The memory allocator in JKA is more of a book-keeping system. `Z_Malloc` delegates to `malloc` so the only difference between using `malloc` directly and using `Z_Malloc` is that the allocation is then tracked by JKA and deleted at end of day, instead of being leaked and being cleaned up by the OS. This also lets us see how much memory we're using for each part of the game. It's not used very well at the moment in all parts of the game, but I'd like to be a bit more strict with that as we progress.

I agree with @smcv that b9251ad and 1db8c6e are commits that are good to merge in.

> As for avoiding std::vector...

As for using standard library functions, I'm not really for or against using them at the moment. On one hand, they will be allocating memory outside of JKA's memory allocator and as I said it would be nice to track all memory that's being allocated. On the other, it's just easier to use them. I think for now it's fine to use in this case, though I would personally use a `std::string` rather than `std::vector`.

ahmedtd:
>>I'm fine with always using the bundled copy, then. I should note, though, that the system copy seemed to work just fine (no stuttering or hiccups from malloc), even on my somewhat anemic laptop.

> The memory allocator in JKA is more of a book-keeping system. Z_Malloc delegates to malloc so the only difference between using malloc directly and using Z_Malloc is that the allocation is then tracked by JKA and deleted at end of day, instead of being leaked and being cleaned up by the OS. This also lets us see how much memory we're using for each part of the game. It's not used very well at the moment in all parts of the game, but I'd like to be a bit more strict with that as we progress.

At the time I wrote that comment, I was under the impression that Z_Malloc was a pooling allocator.  I've since read it a little more closely.

ahmedtd:
It's also important to note that any structure from the stl can have its allocator replaced in a totally standards-compliant way.

ensiform:
In quake3 it is a pooling allocator, in JKA it is not.

ahmedtd:
Merge or no merge? I'll rebase it if you want to merge.

smcv:
I think @ahmedtd's branch is mixing up two orthogonal changes: "build bundled libraries as convenience libraries" (see below for more on that term), and "allow use of a system minizip".

For the bundled libraries that are used, building them as convenience libraries is A Good Thing, IMO (but my opinion doesn't carry much weight here since I don't have commit rights).

Using the system minizip instead of the bundled, modified minizip is a more controversial change, because it alters behaviour (memory might not be freed in some code paths where it previously would have been).

I would suggest that the way to move forward with this would be to prepare a version of the branch that uses convenience libraries when building the bundled libraries, but does not add support for a system minizip; and moving the follow-up, "support for a system minizip", to a separate branch/PR (I don't know whether you can take over #642 from me?) if you still want to pursue it.

> I'm all for compiling more of the libraries as static libs and statically linking, but on Linux it seems to me that dynamically linking is more common. When not using system libraries then using a static lib is a good idea :)

For bundled libraries there are three possible approaches:

Option 1 is to build a proper installable library just like an OS vendor would (usually shared/dynamic, but could be static in principle), and link with it. This is how OpenJK currently deals with the bundled SDL2 (except that it just ships a compiled library rather than compiling it from source). I think this only makes sense for precompiled libraries like the bundled SDL2: if anyone wants to build a proper installable library from source, they should do that separately (from its own source code), install it, and point OpenJK to it as a system library, like I do in Debian.

Option 2 is to build a "convenience library" (that's the Autotools jargon term, I don't know whether CMake has a different term for it). This is a static library (.a) that is never installed, but is only used during compilation of the same source tree it came from. It would make some sense for the bundled libjpeg, libpng and zlib to be convenience libraries, if they are used. It would also make some sense for the modified minizip to be a convenience library that is linked to either the bundled zlib convenience library or the system zlib.

The point of a convenience library is that you only need to compile it once, even if you are linking it into multiple places (for instance compiling zlib once, then linking the convenience library into 4 flavours of the engine, if you enable MP, SP, JK2 and the dedicated server).

Option 3 is to just include the source code in the list of sources for each build product, and live with the fact that you're compiling it 4 times. This is what OpenJK currently does.

xycaleth:
Sorry for leaving this hanging. I'll be able to review this again later on tonight (GMT timezone)

xycaleth:
@ahmedtd 

I would like to keep the minizip code as it is (that is, to continue using `Z_Malloc`) instead of allowing a system-minizip. Keeping the option of tracking memory usage of the minizip code seems pretty important and just recently we've found some issues with other parts of the codebase which doesn't use the memory allocator and was using several hundred megabytes of memory without us realising.

In other words, from the first commit we'd only like to use the changes to build zlib as a 'convenience library'. I've commented inline on your second and third commits but on the whole they look good :)

ahmedtd:
The docfix and buffer overflow have been split into separate pull requests (#701 and #702).  I will rework this PR into just the static library changes.

ahmedtd:
Ok, the ability to use system minizip has been dropped.  The bundled copies of zlib and minizip have been updated to the zlib-1.2.8 release.

ahmedtd:
Sorry, I thought I had fixed all of those.  CACHE variables wouldn't have worked properly anyways.  I've split the bundling configuration into two passes -- in the first, we look at the user's platform and decide what the proper set of default bundled libraries is; in the second, we create the options with the desired defaults.

ahmedtd:
I've given the bundled libpng the same treatment, and I'm also eyeing the bundled libjpeg to do the same thing.

ahmedtd:
Ok, now all the bundled libraries that are included in source form are built as static libs.

ensiform:
We do require a certain version for libpng fwiw. I think 1.6.6 or something  like that. I'll confirm later.

Libjpeg-turbo would be nice too. I think we only support regular right now.

ahmedtd:
libpng and libjpeg are still built from the original source files that were on the master branch.  The only thing that I've done to them is move their public headers into an include/ subdirectory, and set up CMake magic so that the rest of the code can always assume it's using the system version.

libjpeg-turbo has licensing that's more amenable to bundling, I believe.  I figured I should give this branch a rest and let it get merged, though.

Ideally, I think we would not keep the bundled libs in the repo, but just have the user drop a release tarball for, say, libjpeg in the libs/ folder.  Then we'd have a script that would extract the tarball, apply any patches we need, and then build it in static mode.

That's sort of tricky, though, especially to support in an MSVC environment.

smcv:
> Libjpeg-turbo would be nice too. I think we only support regular right now.

The bundled libjpeg is the original "IJG libjpeg", but an external copy of libjpeg-turbo also works fine - that's what I'm using in Debian. You don't need to do anything special to support it.

On Linux and perhaps OS X too, an external libjpeg is what makes most sense anyway, since those platforms can be expected to have reasonable package-management for build-dependencies; so the choice of bundled libraries (if any) should be driven by what you want to happen on Windows.

PR #687, opened 2015-07-29
Use SharedDefines instead of DebugDefines/ReleaseDefines

### LINKED ISSUE #561: CMake 3.0 throws warnings with use of COMPILE_DEFINITION usage per config
With the release of CMake 3.0 use of COMPILE_DEFINITIONS_DEBUG etc is now deprecated therefore throws warnings during CMake build file generation.

smcv:
This greatly simplifies the setup for all the compiled modules, and avoids cmake warnings about policy CMP0043. Since #686 was merged, the SharedDefines are used everywhere.

Fixes #561.


xycaleth:
Looks good! Thanks again for your contributions @smvc :)

PR #688, opened 2015-07-29
fix Windows cross-compilation on Linux using mingw-w64

smcv:
mingw-w64 is a way to compile native 32- or 64-bit Windows executables using gcc/g++, compiling on either Linux (with a cross-compiler that targets mingw-w64) or Windows. The resulting binaries do not depend on a special platform layer like Cygwin: they use the normal Win32 API, with Microsoft's `msvcrt.dll` as their C runtime.

Previous versions of OpenJK needed significant changes for mingw, which is presumably why it isn't currently supported, but current git master can cross-compile under mingw on Linux with only minor modifications: the general theme is that it becomes necessary to distinguish between MSVC-only code, and Windows-only code (which were previously the same thing). These changes hopefully don't break MSVC builds, but I'd appreciate it if someone could test that.

So far, I've only tried compiling 32-bit debug and release builds on Linux. My planned next steps are: test those builds on native Windows; test on Wine; try Win64 builds.

[Travis-CI](https://travis-ci.org/) has a mingw-w64 cross-compiler available, which ioquake3 [uses for snapshot builds](https://github.com/ioquake/ioq3/blob/master/.travis.yml). I'm going to look into setting this up for my own OpenJK branches, and I hope to be able to provide a pull request soon if the OpenJK developers would be interested in doing the same. Unlike the buildbot, Travis-CI can be run automatically for every commit and every pull request, making it easy to detect pull requests from a Linux user that would break the Windows build, or vice versa. You could continue to use OpenJK's current buildbot infrastructure to produce the official snapshots if you prefer that (and it does have the advantage of covering OS X, which needs special setup for Travis-CI).


xycaleth:
A quick look at your changes looks good. I'm a little worried about your changes in 1a9b48184ece to files in the `code/win32/` folder though. Were you seeing compile errors from these files, or did you simply go through all the files? That folder is no longer (or at least, should no longer) be used by any of the executables, so there shouldn't be any need to update them.

We should delete the folder at some point :)

smcv:
> Were you seeing compile errors from these files, or did you simply go through all the files?

I'm not sure; I started the branch a while ago, but I didn't push it, partly because linking failed with the bundled SDL (I think it might only work for MSVC, I'm using a separate SDL binary now) and partly because the diff was a lot larger back then. It's possible that it broke the build when I started, but doesn't any more.

`xinput.h` is what mingw-w64 calls that header, and altering the case should be harmless on Windows, even if it isn't actually any help either :-)

> We should delete the folder at some point :)

One of my ulterior motives in trying to set up a working autobuild environment for mingw is that I can try out changes like "delete `code*/win32` and see what happens" and know that if it builds on all the platforms, it's probably good :-)

smcv:
> the diff was a lot larger back then

I've benefited from some cleanup done by other people within OpenJK, and also some fixes in the mingw-w64 headers (for which I previously had a somewhat nasty workaround).

smcv:
You are right that not all of 1a9b481 is necessary now; the only parts that actually break compilation are the various inclusions of `Windows.h` and `gl/gl.h`, and one `ShlObj.h` in `shared/`. But I'm fairly sure the rest is right, and some of it might be necessary with different compilation options. At worst, it's harmless on case-insensitive OSs.

smcv:
I've updated the branch to stop patching `code*/win32`.

There are a few fixes in code paths which are only taken with `OPENAL` and/or `USE_OPENAL` defined, which we don't (or at least my test builds don't). I've left those in, but split them into their own commit so they're easy to discard if desired.

smcv:
Updated the branch again to be able to use the bundled SDL2 library rather than an external one. It turned out to be a library-order issue.

The change to fix that *should* be harmless on MSVC (but please test!), because the MSVC linker prefers to resolve symbols in the same way gcc would, but unlike gcc it tries to cope with incorrect library order by falling back to libraries earlier in the command line: [The MSVC linker is less sensitive to this issue because it will search all libraries for an unreferenced symbol. Library order still can affect which symbol gets resolved if more than one library have the symbol.](https://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc)

smcv:
> This seems to compile fine when including cmath all the time. Could you change it #include all the time?

Sure. I didn't want to change what happens on MSVC until someone could test there; but if you have now, that's fine.

> Out of curiosity, what happens when this check [for sizeof(void *)] fails?

If the check isn't done (on Windows), then CMAKE_SIZEOF_VOID_P is unset and so doesn't match 8, Architecture is set to x86, and we produce a jagamex86.dll that is actually 64-bit code; and the cmake variable WIN64 isn't set, although that doesn't currently seem to have any practical effect on non-MSVC other than not installing DLLs to `base/`.

I don't think the check can *fail* (as in produce no result), assuming you have a working C compiler. It compiles a simple executable that encodes the size into a string, then looks for the string in the binary (to avoid having to run it, which would fail when cross-compiling).

xycaleth:
Looks good. Thanks smcv.

PR #692, opened 2015-08-01
add Travis-CI continuous integration glue

smcv:
This branch adds the metadata and scripts to build OpenJK for x86_64 Linux (with gcc and clang) and x86 and x86_64 Windows (with the mingw-w64 gcc toolchain), using [Travis-CI](https://travis-ci.org/). x86 Linux is supported but currently disabled, since it takes a few minutes of package installations to get Travis-CI's 3 year old Ubuntu setup into a state where it can produce 32-bit OpenJK binaries, so this build would lag behind the others.

See https://travis-ci.org/smcv/OpenJK for what the build results look like.

It is based on, and requires, my mingw branch (#688) for the mingw-w64 builds. I could disable those and only build for Linux if people really want that, but testing builds for both Windows and Linux seems better. At the moment, any branch in my clone https://github.com/smcv/OpenJK that contains a `/.travis.yml` (meaning just this one right now), and any pull request submitted to my clone (currently none), should get autobuilt by Travis-CI.

If one of the owners of the main OpenJK repository merges my branch, then links their github account to Travis-CI and enables Travis-CI integration for the main repository, the same things will happen. Apart from enabling Travis-CI for `smcv/OpenJK`, the only special setting I made through the web interface was "only build branches that contain .travis.yml" - the rest of the necessary information is in .travis.yml in the branch.

If you don't want Travis-CI hooked up to `JACoders/OpenJK`, it might be useful to merge something like this anyway, so contributors can easily enable it for their cloned repositories (and get a test-build attempted for Windows if they develop on Linux, or vice versa).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ compiler: gcc
 env:
   - ""
   - flavour=Debug
-  - host=i686-linux-gnu
+  #- host=i686-linux-gnu
   - host=i686-w64-mingw32
   - host=x86_64-w64-mingw32
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+language: cpp
+sudo: true
+
+install:
+  - ./tools/travis-install.sh "${host:-native}" "${flavour:-Release}" $opts
+
+script:
+  - ./tools/travis-build.sh "${host:-native}" "${flavour:-Release}" $opts
+
+compiler: gcc
+env:
+  - ""
+  - flavour=Debug
+  - host=i686-linux-gnu
+  - host=i686-w64-mingw32
+  - host=x86_64-w64-mingw32
+
+# One extra build with clang, its warnings are often better
+matrix:
+  include:
+    - compiler: clang
+      env: ""
--- a/tools/travis-build.sh
+++ b/tools/travis-build.sh
@@ -1,0 +1,49 @@
+#!/bin/sh
+
+set -e
+set -x
+
+host="$1"
+flavour="$2"
+shift 2
+
+mkdir deps build
+
+case "${host}" in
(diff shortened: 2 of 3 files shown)


smcv:
Rebased on an updated version of #688, and added a commit to force `CMAKE_SIZEOF_VOID_P` to be set. It looks as though that might happen as a side-effect of something else in the native Linux build, and presumably the native MSVC build too, but it doesn't seem to happen in the cross-build, which might break Win64 detection.

xycaleth:
So I'm not entirely sure what will happen here if #688 gets merged in. I'll hold off this until then.

smcv:
> So I'm not entirely sure what will happen here if #688 gets merged in

This branch contains all the commits from #688, so you can't merge it without also getting the #688 changes (without them, the mingw build would just fail every time). I'll rebase it on the updated version of #688.

xycaleth:
@smcv If you're happy to maintain the Travis metadata files from now on then I can merge this PR in. We just don't want the metadata files there and then for people to find that they don't actually work a few months down the line!

smcv:
> If you're happy to maintain the Travis metadata files from now on then I can merge this PR in

Sounds good to me. If they turn out not to be as useful as I'd hoped, I can send another PR to remove them.

Are you intending to enable Travis integration for JACoders/OpenJK itself, so that all pull requests get autobuilt?

smcv:
> If you're happy to maintain the Travis metadata files from now on then I can merge this PR in

For clarity: yes, I am happy to maintain the Travis metadata files.

I've rebased the branch on current master, and it has built successfully.

xycaleth:
Sorry for the slow reply. I had forgotten to check back.

Everything looks good then if you're happy with it!

PR #693, opened 2015-08-03
make heirarchical date based demo folders

dpadgett:
helps keep the number of folders in each directory at a reasonable level,
otherwise demos accumulate and it gets way out of hand


ensiform:
Looks like you left a lot of commits here that aren't supposed to be here...

dpadgett:
oops, i must have pushed something bad after creating the PR.  i will fix it

dpadgett:
update: it is fixed.

PR #711, opened 2015-09-01
use a worker thread for async io, and use it for server demos

dpadgett:
use c++11 for threading for cross-platform compatibility
add c++11 support to cxx flags


mrwonko:
What problem does this fix?

I'd set `fsh[f].writerThread = nullptr;` in `FS_FCloseAio()` after the delete, initialize it using `std::thread	*writerThread = nullptr;` and `assert(f->writerThread == nullptr)` in `FS_ResetFileHandleData()` to catch any potential leaks, by the way.

I'm not sure how the team feels about C++11 - personally I'd be all for it, but I believe we currently still support VS2010? (I'd be happy to require VS2015 on Windows personally, but not sure how @ensiform et al. think about that.)

xycaleth:
Apparently at some point I said it was fine to use c++11 but not to go crazy with it. Looking at this pull request, I don't particularly like having changes which look completely out of place from the rest of the code...

dpadgett:
the problem is that if a server has a slow hard disk, synchronously saving the server demos can cause lag in game, especially when a new player connects and it has to open a new file.  there is also lag caused on map restart by needing to go through the list of folders and pruning it, but that is not addressed in this patch as it depends on a lot more stuff being thread-safe.

point taken about setting to nullptr, patch updated.

dpadgett:
ping?

Razish:
+1

mrwonko:
+0

I feel uncomfortable merging this since there seems to be disagreement; I don't care much either way. But if this is indeed a performance bottleneck then asynchronous I/O makes sense.

Does this always shut down nicely though? The thread only seems to be joined after writing is done, so it could presumably be terminated when the main thread exits? Admittedly the general lack of RAII makes it somewhat difficult to write correct code...

codemp/qcommon/files.cpp
```diff
@@ -1006,15 +1063,73 @@ void FS_FCloseFile( fileHandle_t f ) {
...
+		}
 	}
-	Com_Memset( &fsh[f], 0, sizeof( fsh[f] ) );
+	FS_ResetFileHandleData( &fsh[f] );
+}
+
+extern void Com_PushEvent( sysEvent_t *event );
+void FS_AsyncWriterThread( fileHandle_t h ) {
+	fileHandleData_t *f = &fsh[h];
+	if ( !FS_CreatePath( f->ospath ) ) {
```
dionrhys: FS_CreatePath should definitely have mutually exclusive access (mutex).

dpadgett: why? all it does is call sys_mkdir.

dpadgett:
hmm yeah, the filesystem code should already close all files on shutdown, but since that can be done asynchronously it should block on joining the thread on final shutdown.

codemp/qcommon/files.cpp
dionrhys: Because it can call Com_Error.

dpadgett: ok fair.  but that just means Com_Error needs to be thread aware right?

ensiform: Com_Error is a noreturn function though with exceptions, can you do that?

mrwonko: Ugh... No threads + exception is tricky... I think if you use std::async the exception is passed along when you .get from the std::future or some such? I'm not sure...

dpadgett: i was thinking i can just make the threads top level function do a try/catch, and propagate the error to the main thread using the event loop.  (that's how it communicates with the main thread now anyway)

mrwonko: That would probably work.

codemp/qcommon/files.cpp
```diff
@@ -263,8 +269,28 @@ typedef struct qfile_us {
...
+			handleAsync(qfalse),
+			writerThread(nullptr),
+			closed(qfalse),
+			fileSize(0),
+			zipFilePos(0),
+			zipFileLen(0),
+			zipFile(qfalse) {
+		ospath[0] = '\0';
+		name[0] = '\0';
+	}
```
mrwonko: [Just use in-class initializers](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c48-prefer-in-class-initializers-to-member-initializers-in-constructors-for-constant-initializers)?

ensiform: Can't you init the char arrays to 0 in the in-class init?

ensiform: FYI

ensiform:
FYI This patch removes MSVC <= 2010

codemp/qcommon/files.cpp
dpadgett: it does not work in vs2013.  http://stackoverflow.com/questions/23900191/error-cannot-specify-explicit-initializer-for-array

dpadgett:
yeah it won't work with <= 2010 now.  you can get 2013 for free though

ensiform:
And 2015, but we still have 2010 support on the repository.

dpadgett:
isn't WinXP end of life anyway?

ensiform:
Yes but there are probably plenty of JA players still using XP.

dpadgett:
Sure but no developers probably. That is what we would lose right?

ensiform:
No, you would lose the players if they can't launch it properly.

dpadgett:
So you're saying xp cannot launch programs compiled in vs2013? Not even if
they have the runtime?

ensiform:
You have to set the toolset but I think @Razish even had to manually set some stuff up to make it work all the time in japp.

mrwonko:
I've successfully compiled OpenJK for XP with VS2015, it's just a matter of supplying `-T vs140_xp` or something like that to CMake and having the platform tools installed. And it should only matter for executables, I think.

xycaleth:
Closing due to inactivity. If you want to submit again, feel free.

PR #716, opened 2015-09-05
Offer opt-out for user tracking functionality (ja_guid)

dionrhys:
### What does this PR address?
Since fd47854, OpenJK sends along a unique ID (`ja_guid`) as part of userinfo to servers, which identifies a player across game sessions and system restarts. It does this by generating a random 2KiB key file (`jakey`) on first run, then sending a hash of that file and the server address when connecting to a server. This is equivalent to [persistent cookies](https://en.wikipedia.org/wiki/HTTP_cookie#Persistent_cookie) on the World Wide Web.

The `ja_guid` tracking can be useful to server owners as they can log and analyse player behaviour over many sessions, even if the player changes their name while playing or before connecting. It can also provide a weak form of correlation to mod authors for logging per-player info such as K/D ratios and combat stats without having to write an accounts system into the mod. _(Please don't implement an admin system using `ja_guid` as the authentication method though)_

### What does this PR do?
This pull request offers players and hosts the choice to opt-out of the default tracking enabled in OpenJK, so that they will not be tracked across game sessions unless they choose to. Players can already identify themselves using their in-game names, and the `ja_guid` method doesn't provide any confidence of authenticity regardless.

The new cvar `cl_enableGuid` is used to toggle tracking functionality. As it's a cvar, players can change the tracking setting at runtime or on the command line. The code comes with tracking enabled by default, as is currently the behaviour in OpenJK.

#### Remarks
For some third-party package distributions, it may be desirable to have a compile-time define for setting whether tracking should be enabled/disabled by default (or not even compiled into the binary at all). I can look into that if it's needed.

_`jakey` files that have already been generated will be left alone._

***

What are your thoughts? :ear:

### DIFF
--- a/codemp/client/cl_main.cpp
+++ b/codemp/client/cl_main.cpp
@@ -91,6 +91,9 @@ cvar_t	*cl_inGameVideo;
 cvar_t	*cl_serverStatusResendTime;
 cvar_t	*cl_framerate;
 
+// cvar to enable sending a "ja_guid" player identifier in userinfo to servers
+// ja_guid is a persistent "cookie" that allows servers to track players across game sessions
+cvar_t	*cl_enableGuid;
 cvar_t	*cl_guidServerUniq;
 
 cvar_t	*cl_autolodscale;
@@ -724,17 +727,32 @@ update cl_guid using QKEY_FILE and optional prefix
 */
 static void CL_UpdateGUID( const char *prefix, int prefix_len )
 {
-	fileHandle_t f;
-	int len;
+	if (cl_enableGuid->integer) {
+		fileHandle_t f;
+		int len;
 
-	len = FS_SV_FOpenFileRead( QKEY_FILE, &f );
-	FS_FCloseFile( f );
+		len = FS_SV_FOpenFileRead( QKEY_FILE, &f );
+		FS_FCloseFile( f );
 
-	if( len != QKEY_SIZE )
-		Cvar_Set( "ja_guid", "" );
-	else
-		Cvar_Set( "ja_guid", Com_MD5File( QKEY_FILE, QKEY_SIZE,
-			prefix, prefix_len ) );
+		// initialize the cvar here in case it's unset or was user-created
+		// while tracking was disabled (removes CVAR_USER_CREATED)
+		Cvar_Get( "ja_guid", "", CVAR_USERINFO | CVAR_ROM );
+
+		if( len != QKEY_SIZE ) {
+			Cvar_Set( "ja_guid", "" );
+		} else {
+			Cvar_Set( "ja_guid", Com_MD5File( QKEY_FILE, QKEY_SIZE,
+				prefix, prefix_len ) );
+		}
+	} else {
+		// Remove the cvar entirely if tracking is disabled
+		uint32_t flags = Cvar_Flags("ja_guid");
+		// keep the cvar if it's user-created, but destroy it otherwise
+		if (flags != CVAR_NONEXISTENT && !(flags & CVAR_USER_CREATED)) {
+			cvar_t *ja_guid = Cvar_Get("ja_guid", "", 0);
+			Cvar_Unset(ja_guid);
+		}
+	}
 }
 
 /*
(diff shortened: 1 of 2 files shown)


ensiform:
I don't really see any reason for this.

Players that don't want the key revealed are only doing it for nefarious purposes anyway and besides IP will still be trackable.  I know of 0 mods that use ja_guid as of yet.  Also with cl_guidServerUniq, the guid is unique per server IP so you cannot track on different servers.

The added cvar also has no purpose when DEDICATED either.

deepy:
Doesn't the ja_guid provide authenticity as long as the server admin is not malicious or set things up incorrectly or the key gets compromised?
And given the size of the playerbase and how GUIDs work, the chance for collisions shouldn't ever be a problem. no?

Razish:
This PR is not about validity/authenticity, but privacy.

dionrhys:
> @ensiform:  I know of 0 mods that use ja_guid as of yet.

OpenJK's gamecode already uses it.

> @ensiform: The added cvar also has no purpose when DEDICATED either.

I can imagine a ja_guid in SERVERINFO on dedicated servers too. That's what I mean by possible server-side tracking.

> @Razish: This PR is not about validity/authenticity, but privacy.

Exactly! If server admins want to use ja_guid as a form of authentication, that's up to them.

`ja_guid` as authentication has about the same reliability as the `password` userinfo variable. Maybe that's good enough authentication for some mods, for others maybe not. `ja_guid` is randomly-generated for each install though, so it's tied to the user's device by default. If they install on a new device, they'd need to copy over the `jakey` file (something no one would realise was necessary until they see their favourite servers don't remember their stats anymore).

ensiform:
It doesn't do anything with it though ?

deepy:
I want to dismiss the privacy concerns.
The guid is unique per server and the server already got your ip. The guid
is exposing no further information about you.

In the case of two people sharing an ip you could argue that it does, but
at that point it's still the same information as if only you'd have
connected

ensiform:
I agree with Cat.

dionrhys:
> @ensiform: Only times I see this being useful would be people wanting to hide under padawan etc.

That's one of the main reasons for this PR. So players have the freedom not to disclose their identities if they don't want to.

In order for players to have privacy (anonymity/pseudonymity) on servers, they need to take care of changing 3 things: userinfo, IP address, and installed PK3s if the server is pure. Changing IP addresses is up to the player, and outside the scope of OpenJK. Installed PK3s are also the responsibility of the players (unique combinations of installed PK3s allow heuristic fingerprinting).

Players control all the userinfo cvars (it's in the name), except for `ja_guid`.. which is a mandatory `CVAR_ROM` variable that they have no control over. Players can remove the `jakey` file, but the game will always generate a new one, and this new identity will persist after the game is closed. By giving players the freedom to choose whether to send the tracker along with them to servers, they have better control over their privacy.

ensiform:
I don't agree with supporting total anonymity tbh because that just supports the hacktivist groups.

xycaleth:
I don't see how it supports people causing trouble. If people are causing trouble then you can just ban their IP address (range).

ensiform:
Because those are the only people who would use such a feature.

dionrhys:
I don't want a tracking GUID sent along with me every time I connect to a server, and I don't think I'm the only player sharing that view. I'm not a trouble-maker. :innocent:

ensiform:
For what purpose does it really bother you? That its put on the server console during connect?

dionrhys:
That it allows server owners to analyse my playing habits and behaviour across game sessions. I value my [privacy](https://en.wikipedia.org/wiki/Privacy).

> **Privacy** (/prɪvəsi/ or /praɪvəsi/; from Latin: privatus) is the ability of an individual or group to seclude themselves, or information about themselves, and thereby express themselves selectively. [...]

deepy:
Re: irc convo,
If players have the option to block guid then I insist that servers gets an
option to block them. The guid system is based on goodwill and all current
stats tracking will break if players report fake guids.

But it's also important being able to distinguish a player opting out and
one lacking a guid, so if they opt out I motion that their guid gets set to
"NOPLZ" so stats tracking can exclude them properly.

dionrhys:
The whole point of this method of unsetting the cvar is to make these clients behave as clients did previously, before any guids were passed over in userinfo. No basejka client will give you a guid of "NOPLZ", so you'll be blocking every client that isn't OpenJK. It would be silly to handle 3 states (valid guid, refused guid, no guid), when the current understanding that some clients don't pass a guid already exists. The fact that the user has *chosen* to disable tracking and not send over a guid is evidence enough that they are not to be tracked using the guid, because it doesn't exist!

Give an option to block non-OpenJK clients if you want, but this cannot be default - it's counter to the project goals. There are no guarantees that any client will report an accurate guid either. Anyone can build the source themselves and send any content they like as a guid.

mrwonko:
> The whole point of this method of unsetting the cvar is to make these clients behave as clients did previously, before any guids were passed over in userinfo. No basejka client will give you a guid of "NOPLZ", so you'll be blocking every client that isn't OpenJK.

Makes sense.

I agree that we should respect the players' privacy. It's effectively a cookie, and many countries now mandate asking the user for permission for using them; the same arguments would apply here, but I'd be okay with not explicitly asking the user, just giving him the documented option of disabling it.

ensiform:
Should this also have a CMake define to disable it completely?

deepy:
Not to argue, but this is nowhere near a cookie, browsers leak PII all over the place (OJK does not) and cookies can follow you between sites identifying you (while the GUID is unique per server).

None of the problems of cookies are relevant.
Though the privacy concerns are.

As for CMake, why not call it GUID?

dionrhys:
I just chose the term 'tracking' because it was the most self-explanatory term I could think of for the cvar. Something like 'enableGuid' or 'enableCookie' could mean anything to the player and OpenJK doesn't support cvar descriptions/help text like the Source engine does.

Indeed the GUID is unique per server, but players could still be tricked by server owners into disabling cl_guidServerUniq so as to be able to impersonate the player on different servers (Which is one of the reasons why you shouldn't create an admin system using the guids for authentication).

I should have internet at home next week so I can look at this more in detail then.

ensiform:
Oh but it does soon.

ensiform:
Though if its such a big issue...  Why has it not been an issue for ET with built in random GUID pre-GPL and ioq3 which both function same way basically.

Razish:
Player count, developer interest.

ensiform:
Pretty sure there are more players playing ET than there are JA.  SP does not count.

dionrhys:
I've renamed the cvar to "cl_enableGuid" and scoped it to cl_main.cpp with a description of what it enables. It's called a persistent cookie in the comment because that's simply what the technical term for it is (though not in the context of a web browser or HTTP here).

I'll refrain from adding a CMake define that omits the ja_guid code entirely as no one's shown interest. It can always be added later if need be.

I'll merge this in soon if there aren't any objections. Asking "why is it needed?" isn't my idea of an objection though! Feel free to :+1: if you agree with the changes too.

PR #738, opened 2015-10-21
Enable modern C++

mrwonko:
I will write some new code that is going to make light use of modern C++ (e.g. for a move-only wrapper around `FS_ListFiles()` that ensures it will be freed); in preparation for this I'd like to enable C++14 support across the board.

Before I start working on any of that I thought I should allow for general discussion; although the only somewhat valid counter point I see is VS2013 support (and even that has support for some C++11/C++14 features). As far as I'm concerned VS2013 support should just be dropped, but let's just agree on allowing clean code (i.e. modern C++) first.


xycaleth:
How well supported would this be for other platforms? Does rpi have a C++14 capable compiler for example. Or older versions of Linux and OSX?

Archangel35757:
I know that I am not a current contributor to this effort, and so my opinion carries little to no weight... 

However, the last version of Visual Studio that I have is 2012 Professional.  And I think you guys should stick to one "official" version of tools until this project is released.  Always moving to the latest version of development tools seems very frustrating, at least to me-- not everybody has a need to move to newer tool versions.  The latest version of Autodesk products (2016) just migrated to VS2012 from VS2010.  

If it is possible to go back in the history and do a fork or pull request that is compatible with VS2010 or VS2012 then I suppose that could work for me and others.  But then those folks would not benefit from changes made using newer tool sets.

Just my two cents.

eezstreet:
To be honest, that's not as big of an issue on other platforms as with VS/MSVC compiler. I would wait a few years tbh. This is now the third time since the project started that the required compiler version would have gone up.


Sent via the Samsung Galaxy S® 5 ACTIVE™, an AT&T 4G LTE smartphone

-------- Original message --------

mrwonko:
> How well supported would this be for other platforms? Does rpi have a C++14 capable compiler for example. Or older versions of Linux and OSX?

I think the most widespread old version of gcc is 4.8, it ships with the latest CentOS and Ubuntu 14.04 LTS (Trusty). It has full C++11 support and barely any C++14. Raspberry Pi seems to ship with the ancient 4.6, but it appears to be easily (?) possible to upgrade to 4.8 or 4.9.

I think the C++11 subset that VS2013 Nov2013CTP supports is thus a good baseline. I'll amend the commit to only use C++11.

An argument could be made for keeping C++11 changes in a separate branch, but I'm not convinced. Outside of unsupported closed source libraries I see no point in staying on VS2012 or below, and you can easily install multiple versions if necessary. We know what happened to the dinosaurs. And the extra hassle of maintaining multiple branches... Meh.

mrwonko:
Apparently MS plans to release a build-only version of VS2015 (without the IDE) in November, so one less reason to reject the upgrade.

Razish:
I think as a modern project it's perfectly valid to require whatever is necessary to support the tools (read: C++14) we want to use. Fortunately all of these tools are freely available.
These minimum requirements only affect developers and build services. If we can satisfy that and have a reasonable use-case for using C++14, I see no reason to hold back.

ensiform:
Why are you clinging to 2012?  2013 and 2015 have community versions that have all the same features as Professional, not the same as Express.

mrwonko:
Okay, C++11 it is then; with the option for C++14 in the future.

ensiform:
What about older server builds?  Idk if all host providers support newer versions of gcc etc so they may be tied to ancient gcc and such?

mrwonko:
FS_AutoFreeFile should be noncopyable though, and the syntax in C++11 for that is much cleaner.

As mentioned I'm thinking of FS_FreeFileList, and I'd like a FS::ListFiles that actually returns a nice wrapper which takes care of that. You'll see a pull request from me to that end soon.

ensiform:
Does it really have to be if the implied intent is that you use it specifically?  This isn't some library where people can use stuff fuck all with a set behaviors being defined etc.

mrwonko:
What, you're arguing because we're not a library we don't need neat code? We are the users, and we will misuse our APIs given the opportunity.

ensiform:
Yes, I am.  I see little purpose in enforcing nocopy in a small code-base with obvious not needing to copy it or move it.

mrwonko:
Especially given that there is no need to copy it (and I disagree that there will never be a need to move it) that should be enforced.

PR #740, opened 2015-10-26
Filesystem RAII

mrwonko:
C++11 RAII wrappers for `FS_FileList()` and `FS_ReadFile()`: `FS::FileList()` and `FS::ReadFile()` with foreach support. Removed `FS_AutoReadFile`, which allowed for misuse.

I think using C++'s namespaces for C++ features makes sense, hence the names.


ensiform:
What about a filehandle class?

mrwonko:
Once I need one or feel like it, that will follow.

PR #741, opened 2015-10-27
move travis-ci from Ubuntu 12.04 to 14.04, and fix the build

smcv:
The master branch [fails to build on travis-ci](https://travis-ci.org/smcv/OpenJK/builds/87619402) because Ubuntu 12.04's compiler does not support `-std=c++11` (#738). This branch switches to travis-ci's new Ubuntu 14.04 infrastructure, and fixes some compiler errors on that platform, resulting in [a successful build again](https://travis-ci.org/smcv/OpenJK/builds/87619298).

See individual commit mesages for more details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: cpp
-sudo: true
+sudo: required
+dist: trusty
 
 install:
   - ./tools/travis-install.sh "${host:-native}" "${flavour:-Release}" $opts
@@ -11,7 +12,7 @@ compiler: gcc
 env:
   - ""
   - flavour=Debug
-  #- host=i686-linux-gnu
+  - host=i686-linux-gnu
   - host=i686-w64-mingw32
   - host=x86_64-w64-mingw32
 
--- a/code/qcommon/q_shared.h
+++ b/code/qcommon/q_shared.h
@@ -109,6 +109,10 @@ along with this program; if not, see <http://www.gnu.org/licenses/>.
 #include <errno.h>
 #include <stddef.h>
 
+#ifdef __cplusplus
+#include <cmath>
+#endif
+
 //Ignore __attribute__ on non-gcc platforms
 #if !defined(__GNUC__) && !defined(__attribute__)
 	#define __attribute__(x)
@@ -777,8 +781,10 @@ inline vec_t VectorNormalize2( const vec3_t v, vec3_t out) {
 int Q_log2(int val);
 
 inline qboolean Q_isnan ( float f ) {
-#ifdef _WIN32
+#ifdef _MSC_VER
 	return _isnan (f);
+#elif defined(__cplusplus)
+        return std::isnan (f);
 #else
 	return isnan (f);
 #endif
--- a/codeJK2/game/Q3_Interface.cpp
+++ b/codeJK2/game/Q3_Interface.cpp
@@ -8613,7 +8613,7 @@ static int Q3_GetString( int entID, int type, const char *name, char **value )
 	case SET_ANIM_BOTH:
 		*value = (char *) Q3_GetAnimBoth( ent );
 
-		if ( VALIDSTRING( value ) == false )
+		if ( VALIDSTRING( *value ) == false )
 			return false;
 
 		break;
--- a/codemp/qcommon/q_math.c
+++ b/codemp/qcommon/q_math.c
@@ -1168,7 +1168,7 @@ float Q_powf ( float x, int y )
 
 qboolean Q_isnan (float f)
 {
-#ifdef _WIN32
+#ifdef _MSC_VER
 	return (qboolean)(_isnan (f) != 0);
 #else
 	return (qboolean)(isnan (f) != 0);
--- a/tools/travis-build.sh
+++ b/tools/travis-build.sh
@@ -42,7 +42,6 @@ set -- -D CMAKE_BUILD_TYPE="$flavour" "$@"
 	-D BuildJK2SPGame=ON \
 	-D BuildJK2SPRdVanilla=ON \
 	-D CMAKE_INSTALL_PREFIX=/prefix \
-	-D CMAKE_VERBOSE_MAKEFILE=ON \
 	"$@" .. )
 make -C build
 make -C build install DESTDIR=$(pwd)/build/DESTDIR
--- a/tools/travis-install.sh
+++ b/tools/travis-install.sh
@@ -7,11 +7,18 @@ host="$1"
 flavour="$2"
 shift 2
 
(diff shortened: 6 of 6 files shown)


CMakeLists.txt
```diff
@@ -212,7 +212,7 @@ elseif (("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU") OR ("${CMAKE_C_COMPILER_ID}" S
 	set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3")
 
 	# enable somewhat modern C++
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
```
deepy: Does this really have to be gnu++11?

smcv: A strictly conforming ISO C math.h wouldn't have `_isnan()`, which is a Microsoft extension. If you specify strict ISO conformance, then you don't get that function declared and the build fails.

I could try making `Q_isnan()` use `isnan()` on mingw-w64 (by changing the conditional for `_isnan()` from `_WIN32` to `_MSC_VER`), but `-std=gnu++11` is the closest C++11 equivalent of how OpenJK was compiled before #738.

smcv: Looks like `isnan()` [isn't declared with mingw-w64](https://travis-ci.org/smcv/OpenJK/jobs/87627720). The compiler error suggests `std::isnan`, but `q_shared.h` needs to be valid C so that `codemp` can use it, so that would require a `#ifdef __cplusplus` or something.

mrwonko: Why can't codemp be C++?

ensiform: I think it was meant as codemp/cgame|game|ui, for compat of legacy mods.

mrwonko: But we can keep the C ABI even if we allow C++ in it...

smcv: > Why can't codemp be C++?

OK, yes, strictly speaking it could.

If there is consensus among the maintainers that you want all the code to be C++, you could rename all the `.c` in codemp to `.cpp`, fix any code that is valid C but not valid C++, and then `q_shared.h` could make use of C++ features... but that seems a much, much larger change than is needed to fix the build.

Razish:
FWIW I compile JA++'s game/cgame/ui modules as C++11 with both legacy and new API working.

CMakeLists.txt
smcv: > The compiler error suggests std::isnan, but q_shared.h needs to be valid C so that codemp can use it, so that would require a #ifdef __cplusplus or something.

This does work as an alternative, although I still prefer the version in this pull request.

https://github.com/smcv/OpenJK/commits/travis-isnan (builds: <https://travis-ci.org/smcv/OpenJK/builds/87767135>)

smcv:
Branch updated, please review:

* Switched to a different way of fixing the `_isnan` issue which @deepy might prefer? If you don't like this one either, please be specific about how you would prefer this compile failure to be fixed, or if you preferred how I fixed it the first time, please pull my `travis-gnu11` branch instead.
* Discarded some JA changes that have been unnecessary since @mrwonko did essentially the same thing in commit 973e9c2. The equivalent change for JK2 (which isn't yet fixed) is still included.
* Added a couple of workarounds for changes to Travis-CI's Ubuntu installation, so that we can exercise compilation for 32-bit Linux again.

Razish:
Just to confirm, this does not break the build for me using Apple LLVM version 7.0.2 (clang-700.1.81)
I have not tested JK2 at runtime. I am building all targets with `cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release -DMakeApplicationBundles=Off -DUseInternalZlib=On -DUseInternalPNG=On -DUseInternalJPEG=On -DBuildJK2SPEngine=On -DBuildJK2SPGame=On -DBuildJK2SPRdVanilla=On ../`

The only warnings for `Q3Interface.cpp` are:
```
/Users/tex/projects/OpenJK/codeJK2/game/Q3_Interface.cpp:127:17: warning: suggest braces around initialization of subobject [-Wmissing-braces]
        stringIDExpand("", BSET_INVALID),
        ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
        {
/Users/tex/projects/OpenJK/codeJK2/game/Q3_Interface.cpp:82:38: note: expanded from macro 'stringIDExpand'
#define stringIDExpand(str, strEnum)    str, strEnum, ENUM2STRING(strEnum)
                                        ^
/Users/tex/projects/OpenJK/codeJK2/game/Q3_Interface.cpp:128:2: warning: suggest braces around initialization of subobject [-Wmissing-braces]
        "",                             -1,
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        {                                 }
/Users/tex/projects/OpenJK/codeJK2/game/Q3_Interface.cpp:133:2: warning: suggest braces around initialization of subobject [-Wmissing-braces]
        "NULL",WP_NONE,
        ^~~~~~~~~~~~~~
        {             }
/Users/tex/projects/OpenJK/codeJK2/game/Q3_Interface.cpp:2306:2: warning: suggest braces around initialization of subobject [-Wmissing-braces]
        "", TEAM_FREE,
        ^~~~~~~~~~~~~
        {            }
```

smcv:
> I misinterpreted your commit message as "this commit will break the build with Clang" :)

Right, no. The "this" in "This breaks the build" meant "the bug I'm fixing" :-)

> if we can confirm it working on other platforms

[Compile-tested successfully](https://travis-ci.org/smcv/OpenJK/builds/115593519) with: gcc targeting x86 and x86_64 Linux; gcc (mingw) targeting x86 and x86_64 Windows; clang targeting x86_64 Linux. In addition, you've compiled with clang targeting whatever your Mac is, and AppVeyor has compiled with MSVC targeting x86 and x86_64 Windows.

What I tested in the sense of running the game is x86_64 Linux, and x86 mingw binaries on 64-bit Windows 10.

smcv:
I also have scan-build *almost* working on travis-ci, but I'm having some trouble combining scan-build with mingw, and it takes 40 minutes per configuration, compared with 10-15 minutes for just compiling. I don't think this is something I'd want to run for every pull request.

Razish:
That's okay, we can investigate automated static analysis options later if at all. I'm happy to occasionally run offline analysis.
The combination of MSVC+GCC+Clang with good configurations will provide plenty of warnings to look at.

PR #743, opened 2015-10-31
Add GSL

mrwonko:
In some code I'm currently writing I make use of the Guideline Support Library (GSL), so I thought I'd start by discussing using that in OpenJK.

The [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines) are a set of rules for creating more robust software; to this end a couple of new classes such as `not_null<T*>` and `owner<T*>` are introduced as the GSL. (By way of an example: The point of `owner` is to clarify that this pointer needs to be freed (or handed off somewhere) to prevent leaks.)

Critically these new types are made with legacy C ABIs in mind; as such e.g. `owner<T>`is really just `T` and its sole point is to aid automatic verification using tools; the first one, Microsoft's checker, will be released soon alongside Visual Studio Update 1, together with their reference implementation of the GSL.

However Microsoft's GSL implementation (whose preliminary version [is on Github](https://github.com/Microsoft/GSL/)) only works in GCC 5.1+, and we still want to support 4.8. As such I've chosen to include [GSL Lite](https://github.com/martinmoene/gsl-lite), a backwards compatible implementation that is mostly identical feature-wise. Eventually the GSL implementation to use could be dynamically chosen in CMake; Microsoft's GSL checker may require usage of their implementation.

But even outside of automatic checking the GSL has some useful features; in particular it contains `array_view`/`string_view`, non-owning array views which consist of a start and end pointer, such that they know the array's size and prevent past-the-end access while providing constant-time length determination. I'm currently rewriting GenericParser2's `GetToken()` using them and it makes for much nicer code than null-terminated strings; for example I can just return a `string_view` into the source buffer instead of having to copy the current token so I can null-terminate it.

So we should totally start using it in OpenJK.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,6 +151,7 @@ set(MPDir "${CMAKE_SOURCE_DIR}/codemp")
 set(JK2SPDir "${CMAKE_SOURCE_DIR}/codeJK2")
 set(SharedDir ${CMAKE_SOURCE_DIR}/shared)
 set(OpenJKLibDir "${CMAKE_SOURCE_DIR}/lib")
+set(GSLIncludeDirectory "${OpenJKLibDir}/gsl-lite/include")
 
 include(InstallConfig)
 
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -42,7 +42,7 @@ if(BuildSPEngine OR BuildJK2SPEngine)
 		${SPEngineIncludeDirectories}
 		${SharedDir}
 		${SPDir}
-		"${OpenJKLibDir}/gsl-lite/include")
+		"${GSLIncludeDirectory}")
 
 	#    Dependencies
 	# OpenAL (is optionally included for Windows)
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -41,7 +41,8 @@ if(BuildSPEngine OR BuildJK2SPEngine)
 	set(SPEngineIncludeDirectories
 		${SPEngineIncludeDirectories}
 		${SharedDir}
-		${SPDir})
+		${SPDir}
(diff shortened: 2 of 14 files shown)


ensiform:
Is this something that we really need?  I don't envision OJK taking off as some modern C++ codebase unless you plan on rewriting it all yourself.

mrwonko:
This is mostly for any new code I write; but I intend to modernize everything I use from such new code (hence the recent filesystem RAII & zone allocator code).

In particular I'm currently planning on expanding the Dynamic Music System code (which loads `ext_data/dms.dat`) to allow for loading of addition files in `ext_data/mds/*.dat`; to this end I will rewrite and modernize it, and for that I need file system access and the GenericParser2, which I'm currently rewriting/modernizing using gsl's `string_view`, which is a wonderful alternative to null-terminated strings.

If eventually the entire code base was modernized, that'd be wonderful though. Given how much I enjoy cleaning it up so far, maybe we'll see that happen over the next years.

ensiform:
Could sstring be replaced with a std::array + string_view ?  The listed reason seems to not be totally correct as to why its not just a std::string.  It could just be made into a std::string but there's a size restriction on the instances used (MAX_QPATH=64)

mrwonko:
What's sstring?

ensiform:
Its one of their classes used for places with std::map in the codebase so it says in the comments.  But its basically just a char[64] with some functions to be stl compat.

mrwonko:
Yeah that would be possible I suppose.

ensiform:
StringEd code could probably do with using string_view as well probably.

mrwonko:
Pretty much everywhere a null-terminated string is used, a string_view is better. We're somewhat limited by the ABI though.

PR #748, opened 2015-11-08
[SP] generic parser 2 cleanup

mrwonko:
I've greatly simplified & modernized the Generic Parser 2 in SP; this partially extends to the code using it, in particular the FX Parsing, which also got a first simplification pass (but there's still potential for much more).

The parser is used by FX and DMS parsing; at a glance both effects and music seem to work properly with the new system.

In particular this:
* Gets rid of the unused C interface to the parser
* Removes the unused & unfinished write functionality
* Removes the TextPool (to which all parsed strings were copied) in favor of using string views into the contents of the parsed file, thus removing the need for copies altogether (unless you need a string to be null-terminated)
  - as part of this a bunch of functions were made to use string views instead of null-terminated strings
* properly handles trailing whitespace (in theory; I have yet to verify this)
* removes some of the code duplication in FX Parsing; but it's still kind of a mess, as is the fx handling itself
* fixes a couple of small bugs like accepting a trailing '}' in the generic parser 2
* only includes genericparser2.h where it's needed

Initially this was in preparation of a rewrite of the DMS (dynamic music system) code to allow for the loading of additional files; I'm not sure when that will happen though.

~~The commits marked as [WIP] don't compile; and the commit dates are misleading, I've been rebasing.~~

### DIFF
--- a/code/cgame/FxScheduler.cpp
+++ b/code/cgame/FxScheduler.cpp
@@ -443,7 +443,7 @@ int CFxScheduler::RegisterEffect( const char *file, bool bHasCorrectPath /*= fal
 	char			*data;
 	char			temp[MAX_QPATH];
 	const char		*pfile;
-	char			*bufParse = 0;
+	const char		*bufParse = 0;
 
 	if (bHasCorrectPath)
 	{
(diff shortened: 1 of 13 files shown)


mrwonko:
Squashed the commits together so they all build.

PR #752, opened 2015-11-21
CMakeModules: MSVC_RUNTIME

eternalcodes:
Configure projects to use statically-linked runtime (/MT and /MTd) instead of the default dynamically-linked runtime (/MD and /MDd). Run cmake with "-D MSVC_RUNTIME=static" to switch.


ensiform:
Why static?

eternalcodes:
So that people won't have to install the VC Redistributable Packages. It's annoying when you release a mod and someone is having trouble running it because of missing DLLs.

Why not static?

ensiform:
You're supposed to distribute it. And having static means it must be recompiled of there's updates to the run time but not a new visual studio.

eternalcodes:
There's only a pk3 being distributed, which is sometimes automatically downloaded from the server.

eternalcodes:
This commit only gives people the option to generate projects with statically-linked run-time. You can still generate projects normally with no changes to the .bat files. The default is to use dynamically-linked run-time. It was a pain for me to manually convert all projects to use static every time I generated new project files. There may be others who wish to do the same. With this commit, it makes it easier for them to do it.

PR #779, opened 2016-01-01
Make clear where to put the executables and what they are for

jubalh:
When unzipping I got two subfolders that were not mentioned in the
documentation. I was not aware that the folder structure had to look
like this and had to find out on IRC about it. Let's make it more clear
for others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,9 @@ Installing and running OpenJK:
 
 1. [Download the latest build](http://builds.openjk.org) for your operating system.
 2. Extract the contents of the file into the Jedi Academy `GameData/` folder. For Steam users, this will be in `<Steam Folder>/steamapps/common/Jedi Academy/GameData`.
+*Note:* The archive might unzip into `install/JediAcademy/` and in there is all the data. Actually *this* data should be in `GameData/` folder.
 3. Run `openjk.x86.exe` (Windows), `openjk.i386` (Linux 32-bit), `openjk.x86_64` (Linux 64-bit) or the `OpenJK` application (OS X), depending on your operating system.
+The executable `openjk` is the multiplayer mode, `openjk_sp` is the single player/story mode and `openjkded` is a standalone server for multiplayer games.
 
 
 **OS X Instructions**


ensiform:
Under what OS under what conditions?  This isn't exactly a release state but yes.

jubalh:
Under openSUSE Tumbleweed (Linux) extracting with: tar xfv openjk-2015-11-30-c8c64c4b-linux-64.tar.gz

Anyways I think the information added is universal. It's good to know that the binary needs to be directly in the GameData folder.

marvinhagemeister:
+1 I was initially unsure as well where to put the binaries under osx. A short note in the docs would be awesome!

jubalh:
@ensiform  anything you are not happy about with that change? :-)

ensiform:
The subfolder GameData isn't technically necessary for Linux or Mac.

Razish:
With no Linux installer, there's no standard/provided `GameData` folder.
I don't think there's a GameData folder with any disc copy or App Store version of JA on Mac?
Steam and Windows installers are the only ones to contain `GameData` AFAIK?

ensiform:
In fact on linux it should not be used since most servers don't have that directory hierarchy.

jubalh:
But funny thing about that is that the info about `GameData` folder was in that README even before I added the additional information. So there clearly should something be changed

jubalh:
Well two of my friends who wanted to install the game had the some unclarity I had. They called me and I explained them. Would have been easier if this would have been in the readme already..

Razish:
I've had issues trying to explain to Mac users where to put things and what GameData is. This issue is bigger than (and not caused by) OpenJK.
How do we resolve this? A standardised directory + structure would be best, but we can't make assumptions about user configurations. Most installation targets are different. Maybe the best way is to provide a table listing common paths and describing the agreed-upon "best location" for readable + writable directories. May need some Mac and Linux users; we should cover App Store, Steam and previous installations in the case of no native installers

jubalh:
Sounds like a lot to figure out.
My setup is like this:
I have the CD-Rom Version of the game and run Linux.

With this one I can help, about the others I have no idea.

Still my suggestion is to add the information I proposed already to make it easier for users right now. And have a best-case-solution later.

Razish:
Just looked over it again, I don't think that install prefix should be there at all.
On Linux with `-DCMAKE_INSTALL_PREFIX=/home/tex/games` I get:
```
Install the project...
-- Install configuration: "RelWithDebInfo"
-- Installing: /home/tex/games/JediAcademy/openjk.x86_64
```

On Mac with `-DCMAKE_INSTALL_PREFIX=/Users/tex/games` I get:
```
Install the project...
-- Install configuration: "RelWithDebInfo"
-- Installing: /Users/tex/games/JediAcademy/openjk.x86_64
```

On Windows I have `-DCMAKE_INSTALL_PREFIX=D:/Games` and a `D:/Games/JediAcademy` folder "symlinked" to `steamapps/common/Jedi Academy/GameData` - but you could just as easily link it to your existing installation's GameData if you don't use Steam.
My Windows setup isn't ideal because it involves things most users have never done before.
If we change the Windows build output to match common paths on Windows (i.e. GameData) then it's broken again for Steam installations or (some?) Mac installations.

...what I'm saying is, the instructions in this PR mention an `install` parent folder, which should not exist. ~It's possible this has already been fixed, it's possible this is just a buildbot misconfiguration where it's zipping and uploading the wrong directory~
I believe the `install` folder should exist with the buildbot configuration and contain `JediAcademy` as-well as `JediOutcast` in the future. That sounds reasonable.

That being said, how do others feel about this change now?

Razish:
Ping

bartrpe:
Made renundant with the merge of: https://github.com/JACoders/OpenJK/pull/1171 

@Razish @ensiform plz lock

PR #793, opened 2016-01-18
Removed unused files and namespace std;

dwrj87:
Nothing from the characters.h was used, and it wasn't even in one of the CMake files, and removed some using namespace std and put in std::


xycaleth:
All looks good to me! Thanks for the pull request :)

ensiform:
While skimming this I noticed there are some iterators using postfix rather than prefix.  Should these get fixed in a separate?

dwrj87:
@ensiform I'll fix them later on if you want

PR #794, opened 2016-01-20
Stripped a load of unused trek code

dwrj87:


Archangel35757:
What if mods want to use some of that functionality?  Just because something isn't used by JKA doesn't mean you should strip things out.

Razish:
None of it was functional, so mod authors would have had to reimplement it anyway, which they still can.

ensiform:
Yes, as @Razish said, all of it was commented out and not used.

PR #799, opened 2016-01-22
add cl_drawRecording

deathsythe47:
Hide ugly "RECORDING xyz.dm_26" message on screen. Useful for players who record every game and don't need to see this.

### DIFF
--- a/codemp/client/cl_main.cpp
+++ b/codemp/client/cl_main.cpp
@@ -102,6 +102,8 @@ cvar_t	*cl_consoleKeys;
 
 cvar_t  *cl_lanForcePackets;
 
+cvar_t	*cl_drawRecording;
+
 vec3_t cl_windVec;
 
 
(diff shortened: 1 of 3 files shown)


Razish:
Sounds good, until we redo it into a simple `o REC` notification ala jaMME

entdark:
https://github.com/mvdevs/jk2mv/commit/25c3b88d3da571ac072b85c6b85c2ddf355ffc0d
try this

ensiform:
Can't use the recording image part of course. @ent

entdark:
@ensiform Why? I drew it myself, I let you do what ever you want with it, even those dirty thoughts in your mind. ;P

ensiform:
Assets aren't distributed with ojk.

xycaleth:
I don't see any reason why assets we create ourselves can't be distributed. We already distribute the ojk icon.

ensiform:
Well where would you load it from so that it always gets used regardless of mod?

deathsythe47:
The ojk icon is necessary and helpful to distinguish ojk from base in the operating system.

However, I don't see why this trivial engine tweak requires custom images. Original or not, including dank assets for every minor tweak seems like a slippery slope toward out-of-scope land. Why not make a graphical speedometer for cg_drawfps, or a stopwatch image for cg_drawtimer?

ensiform:
I agree with death.

ensiform:
Hm GH says this is still good, is everyone okay with merging ?

PR #800, opened 2016-01-25
[README] Fix OS X instructions

isair:
Latest builds of OpenJK seem to search for SDL in homebrew installation locations. They actually crash on launch unless SDL is installed via homebrew.


Razish:
+1, homebrew is a good way to standardise packages and instructions should match this.

Issue #780, opened 2016-01-02
Loading Bug


TheCherry:
Hi,

I have a loading bug in OpenJK.

Tested Versions:
openjk-2015-11-30-c8c64c4b-linux
openjk-2015-11-30-c8c64c4b-linux-64

System:
Linux Mint 17.1 - 64bit

Problem:
If I loading a game (from ingame) some objects don't interact correct. As example the tree at start cant be break. Here a video: https://youtu.be/DFTr8XYZf88

(The crash at the end of the video is course I try to load a save file from 64bit version in the 32bit version.)

ensiform:
Saves are not cross arch compatible.

TheCherry:
Yes thought that. Just wanted to explain, why its crash at the end of the video.

yomli:
Same here. Seems related to issue #257. 
Version 2016-01-04-db94c70b-linux-64 on Linux Mint 17.3.

My workaround : restart and load.

xycaleth:
I believe this is a duplicate.of #257.and #295. I fixed it in e8fc63788, but it was potentially broken again in bace05c9b3d when the `-fvisibility=hidden -fvisibility-inlines-hidden` were removed for GCC builds...

PR #811, opened 2016-03-16
Changed FxPrimitives .cpp and header

gort818:
Add the inline tag to function declaration in the  header file, and removed inline from cpp file. Fixes an issue in linux where certain events will not reload after death. Tested three times on different linux machines, built and ran.  Have not tried on windows yet. Fixes this issue #780


ensiform:
Would prefer you to remove the merge commits before accepting.

xycaleth:
Looks like I commented in the wrong place earlier:
https://github.com/JACoders/OpenJK/issues/780#issuecomment-197585021

mrwonko:
I've rebased and squashed these commits as 022ffc7.

PR #815, opened 2016-03-19
Upgrade to SDL 2.0.4

mrwonko:
Let's see if this causes any trouble with the automatic builds, though I suppose not.


mrwonko:
Apparently we no longer need different .libs for different MSVC Runtimes (and thus VS versions and toolsets) because SDL no longer uses the C runtime; but I'd like someone who's still using VS2013 (if we still support that) to verify that.

PR #816, opened 2016-03-20
Adds the final folders for buildbot and adds argument to make.bat for specifying source location.

deepy:


tools/builds/make.bat
```diff
@@ -1 +1,5 @@
-cmake -G "Visual Studio 12" -D CMAKE_INSTALL_PREFIX=install -D CMAKE_BUILD_TYPE=Release "--build build" .
+set BUILDDIR=%1
+IF [%BUILDDIR%] == [] (
+    set BUILDDIR=.
+)
```
mrwonko: How about a simple `mkdir %BUILDDIR%`?

deepy: Thinking about it, builddir is probably the wrong name since it's supposed to point to the source (build the thing in this directory, not build the thing using this directory)

smcv: The conventional naming in at least Autotools is that you cd into the `builddir` to build what's in the `srcdir`. The `srcdir` is the one that contains `.git`, and if you're doing something like `cd build && cmake (some options) ..`, then `build` is the `builddir`.

If there's no reason to do something different, I'd suggest using the Autotools naming: it's what random open-source developers are likely to be familiar with.

deepy: BUILDDIR has now been replaced with SRCDIR to better convey what it's used for :-)

xycaleth:
We don't have a Windows build bot anymore!

PR #819, opened 2016-03-25
ogl improvements

ouned:
noticed two things while porting new SDL code to JK2MV:
- if the opengl context can't be generated it still goes through GLimp_SetMode and returns RSERR_OK.
- no option to allow software GL renderers (OSX VM's have no GPU acceleration, only option you have is to use Apple's software renderer, which is what I'm doing)

### DIFF
--- a/shared/sdl/sdl_window.cpp
+++ b/shared/sdl/sdl_window.cpp
@@ -619,6 +619,11 @@ static rserr_t GLimp_SetMode(glconfig_t *glConfig, const windowDesc_t *windowDes
 					glConfig->colorBits, glConfig->depthBits, glConfig->stencilBits );
 			break;
 		}
+
+		if (opengl_context == NULL) {
+			SDL_FreeSurface(icon);
+			return RSERR_UNKNOWN;
+		}
 	}
 	else
 	{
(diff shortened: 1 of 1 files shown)


shared/sdl/sdl_window.cpp
```diff
@@ -723,6 +729,7 @@ window_t WIN_Init( const windowDesc_t *windowDesc, glconfig_t *glConfig )
 	Cmd_AddCommand("minimize", GLimp_Minimize);
 
 	r_sdlDriver			= Cvar_Get( "r_sdlDriver",			"",			CVAR_ROM );
+	r_allowsoftwaregl	= Cvar_Get( "r_allowsoftwaregl",	"0",		CVAR_ARCHIVE|CVAR_LATCH );
```
xycaleth: Can you rename the cvar and variable to use camelCase? e.g. `r_allowSoftwareGL` Just to be consistent with everything else.

ouned: here you go

xycaleth:
Looks good! :)

ensiform:
Ideally we don't really want software rendering on normal PC's though.

leilei-:
and I think an nvidia driver bug on linux doesn't return accelerated visual for it to start up... or some reason like that why ioq3 removed r_allowSoftwareGL's restrictions so long ago

ouned:
No need to worry guys
Its turned off by default

ensiform:
That's not really the worry.  It means we have to maintain support for it if it's going to be a toggle.

PR #827, opened 2016-04-15
R_PrintLongString bugfix

eternalcodes:
Words are no longer cut off allowing console to word wrap properly.

### DIFF
--- a/code/rd-vanilla/tr_init.cpp
+++ b/code/rd-vanilla/tr_init.cpp
@@ -1186,16 +1186,28 @@ Workaround for ri.Printf's 1024 characters buffer limit.
 
 void R_PrintLongString(const char *string) {
 	char buffer[1024];
-	const char *p;
+	const char *p, *s;
 	int size = strlen(string);
+	int copySize, bufferSize;
 
+	bufferSize = sizeof(buffer);
 	p = string;
-	while(size > 0)
+	while (size > 0)
 	{
-		Q_strncpyz(buffer, p, sizeof (buffer) );
-		Com_Printf( "%s", buffer );
-		p += 1023;
-		size -= 1023;
+		s = p + bufferSize - 1;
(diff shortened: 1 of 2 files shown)


ensiform:
Can you show instances of comparison?  Where it would have cut off vs now?  I'm happy to merge it I just am curious to see.

Razish:
This looks like particularly long strings with no spaces would have severe issues (e.g. infinite loop, OOB access)
Has it been properly tested? Not just with typical data, but targeted unit testing?

xycaleth:
@Razish How do you unit test something which has dependencies on Com_Printf?

Razish:
You use a loose definition of "unit testing" and/or stub it out :p semantic pedantic!

Are we happy with the update?

code/rd-vanilla/tr_init.cpp
```diff
@@ -1186,16 +1186,28 @@ Workaround for ri.Printf's 1024 characters buffer limit.
...
+	bufferSize = sizeof(buffer);
 	p = string;
-	while(size > 0)
+	while (size > 0)
 	{
-		Q_strncpyz(buffer, p, sizeof (buffer) );
-		Com_Printf( "%s", buffer );
-		p += 1023;
-		size -= 1023;
+		s = p + bufferSize - 1;
```
dionrhys: This line of code is navigating 1023 bytes past the pointer p and then dereferencing that address on the next line. In effect this means that at the last iteration of the loop, it will **always** read memory past the end of the original string - thus a buffer over-read.

dionrhys:
I've just pushed 84978e7 to fix this issue. You may want to make use of this version in your own fork to prevent the buffer over-run problem (assuming no one finds one in mine too!).

mrwonko:
> How do you unit test something which has dependencies on Com_Printf?

Simply supply Com_Printf as a template argument.

PR #828, opened 2016-04-21
CL_ConsolePrint Improvements

eternalcodes:
Moved cl_noprint if statement up
Faster word length counter by avoiding recounting word length on same word

### DIFF
--- a/codemp/client/cl_console.cpp
+++ b/codemp/client/cl_console.cpp
@@ -442,6 +442,11 @@ void CL_ConsolePrint( const char *txt) {
 	qboolean skipnotify = qfalse;		// NERVE - SMF
 	int prev;							// NERVE - SMF
 
+	// for some demos we don't want to ever show anything on the console
+	if (cl_noprint && cl_noprint->integer) {
+		return;
+	}
+
 	// TTimo - prefix for text that shows up in console but not in notify
 	// backported from RTCW
 	if ( !Q_strncmp( txt, "[skipnotify]", 12 ) ) {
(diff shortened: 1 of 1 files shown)


codemp/client/cl_console.cpp
```diff
@@ -478,19 +479,22 @@ void CL_ConsolePrint( const char *txt) {
 		}
 
 		// count word length
-		for (l=0 ; l< con.linewidth ; l++) {
-			if ( txt[l] <= ' ') {
-				break;
+		if ((l < 0 ? l = 0, true : false) || (l + 1 == con.linewidth)) {
```
xycaleth: This is pretty unreadable..

eternalcodes: If length is ever less than 0, set it to 0 and return true, else return false. So if it's less than 0, it'll count word length but start with txt[0].
(l + 1 == con.linewidth) this is to continue counting word length since the while loop stops when the length is == con.linewidth

ensiform:
There's really no benefit to moving cl_noprint up.  Also, cl_noprint is basically useless if you use con_notifytime < 0 since cl_noprint stops printing all together as opposed to just hiding the notify box.

eternalcodes:
If cl_noprint is 1, then it'll return sooner than checking for [skipnotify] and chat messages.

codemp/client/cl_console.cpp
xycaleth: I understand what it does, but code that takes effort to understand isn't great, especially when it doesn't add any benefit over writing it out normally.

eternalcodes: I just wanted to check for l < 0 once instead of twice.

dionrhys:
I feel this code is pretty unreadable. The added complexity outweighs the negligible performance improvement that will be had here. In the absolute worst case scenario where there's a word that spans the entire length of the console width (76), the for loop will execute 2926 times (76 + 75 + 74 + ... + 1). This won't show up on any profiler - the CPU will likely have the entire line of text in cache from the first iteration of the outer loop, meaning it won't even need to load new data from RAM for subsequent iterations.

If you've found there are real performance concerns with the original code, please create a test case so we can reproduce this ourselves so we can look at changing things. For now, I'm closing this issue.

Issue #824, opened 2016-03-30
Exceptions across dll boundaries are unsafe


mrwonko:
Com_Error throws an exception, which may not pass correctly back through mod code, especially not when it's built with a different compiler; maybe this can be addressed in the non legacy api. Each api call would have to return an error code, I suppose, and then throw a new exception.

I'll investigate.

xycaleth:
#254 probably has some relevant information

Issue #739, opened 2015-10-25
Serialization for Save games


ensiform:
Save games should be serialized if possible to fix portability issues with saves.  Not sure if its possible to retain the format though.  Serialization doesn't seem to go well with how they currently handle things with string limitations and also chunk ids.  Also not sure how to serialize across dll boundaries since sv_savegame and g_savegame are both needed to handle the save contents.  And avoiding C++ objects across the dll boundary is required.

mrwonko:
Ugh, this sounds like a lot of work.

Compat with existing Windows saves should be possible though, if I'm not mistaken.

ensiform:
I don't see how that works with chunkids though.

ensiform:
If new format: https://google.github.io/flatbuffers/ ?

mrwonko:
Maybe. But if possible, no new format.

ensiform:
Should saves be something that needs to be done prior to a release?  I kind of think so.

mrwonko:
Maybe #434 can also be fixed as part of this; we'll see.

mrwonko:
Try fixing #757 as part of this, too.

ensiform:
Not sure if the new code qualifies as serialization or maybe it does but now that #878 is done can this be closed?

bb0b2f0a541117aba77fc7b59ee51defda99a5b6

smcv:
> Not sure if the new code qualifies as serialization

I think it does, but one issue that still remains is that it writes numbers in the CPU's native endianness. So you can now have savegames that are portable between 32-bit and 64-bit x86, but not between 32-bit x86 and 32-bit PowerPC.

Because savegames seem to start with a 32-bit "magic number", it would be possible to check whether the magic number is in native or opposite-of-native endianness, and set a flag in the reader object; every time a multi-byte number is read, the reader object would have to byteswap it, and every time an array of numbers is read, the reader object would have to byteswap each member.

bibendovsky:
Please do not close the issue, I'll fix the endianness stuff.

smcv:
Loading ordinary integers etc. (for example array lengths and the FPSL chunk) should be something the template code can do. I noticed while fixing #902 that the Icarus stuff saves/loads its own formatted binary blob, which won't be affected by generic endianness-swapping - but code/icarus/BlockStream.cpp uses LittleLong() so it looks as though someone already thought about endianness there. Any other arbitrary binary blobs will need special care.

[...]

PR #848, opened 2016-06-28
Fix loading/saving saved games for 64-bit mode

bibendovsky:
Made saved games compatible with 32- and 64-bit platforms.
Added an additional class to each save/load affected one to handle proper reading/saving data.


ensiform:
Excellent!  You must have been working in this for quite while.

bibendovsky:
Yes, those wrappers took some time.
I noticed that Jedi Outcast saved games not compatible with original ones. Should be this fixed too?

xycaleth:
I've got a few comments on this which I'll make when I finish work.

xycaleth:
And yes, if you think you can keep/fix compatibility, please do :)

ensiform:
Fixing the JK2 saves might be a bit more work though since you have wrappers to specify what goes where, you should be able to diff the original source to find the problems with some of the shared code.

xycaleth:
Right, so I'm not going to code review 9000 lines of code. I do have some general observations though:

1. I like that each struct that gets written has its own import and export method.
2. I don't like that each struct effectively has a 'mirror'. It just means you have to remember to update another struct.
3. I *really* don't like the over-use of "modern C++" for what effectively boils down to reading/writing data to/from a file. Building on your import/export methods, I would much prefer to see something like this:
```c++
void export( some_kind_of_stream& s ) {
  s.write<int32_t>(legsAnim);
  s.write<int32_t>(torsoAnim);
  s.write<float>(animTimer);
  s.write<vec3_t>(position);
}

void import( some_kind_of_stream& s ) {
  legsAnim = static_cast<int>(s.read<int32_t>());
  torsoAnim = static_cast<int>(s.read<int32_t>());
  animTimer = s.read<float>();
  s.read<vec3_t>(position); // this is awkward, i know
}
```

ensiform:
Serialization would be nice

#739

mrwonko:
I'll take a look at this at the weekend, I'm curious.

bibendovsky:
I did not want to touch game interfaces so made a set of wrappers. Serialization is better anyway.

bibendovsky:
Update 1.
Replaced wrappers with a specialized class.
In the next update I plan to fix Jedi Outcast saved games compatibility with the original game.

ensiform:
Still seems a bit over-engineered for the sake of C++.  I think @Razish agrees too?  Anyone else?  Keep in mind I don't think the official builds have been pushed to 2015 from 2013 yet so some newer C++ features may still be unavailable.

bibendovsky:
> Still seems a bit over-engineered for the sake of C++.

Well, almost all projects in solution compiles with C++ compiler.

> Keep in mind I don't think the official builds have been pushed to VS2015 from 2013 yet so some newer C++ features may still be unavailable.

System requirements for mentioned studios are identical. What prevents from upgrading? Also I don't use languages features beyond C++11.

> Also there's an ABI issue by having a non pod class type across boundaries.

Not POD but interface. POD will return us to the wrappers.

ensiform:
Sorry but not everyone upgrades.  And 2013 doesn't even support constexpr properly.

shared/qcommon/ojk_saved_game.cpp
```diff
@@ -0,0 +1,793 @@
...
+{
+    close();
+
+
+    const auto&& file_path = generate_path(
+        base_file_name);
+
+    auto is_succeed = true;
+
+    if (is_succeed)
```
xycaleth: Redundant check?

shared/qcommon/ojk_saved_game.cpp
```diff
@@ -0,0 +1,793 @@
...
+
+bool SavedGame::is_writable() const
+{
+    return is_writable_;
+}
+
+void SavedGame::read_chunk(
+    const SavedGame::ChunkId chunk_id)
+{
+    io_buffer_offset_ = 0;
```
xycaleth: Add a check here to test if `file_handle_` is valid.

xycaleth:
I've had a bit of a closer look and the overall design looks good. I'll make a few comments about the implementation and some parts of the design:

1. Throwing exceptions from ISavedGame is a no go. Exceptions (of user-defined types) don't work across DLL/shared library boundaries as the binary and library might be compiled with different compilers, and different compilers deal with exceptions differently and it all becomes a huge mess. We've already had big issues with the original JKA code doing this which I had to patch up.

2. You have a number of `if (is_preview_mode_)` conditions wrapped in `#if 0`. Could you use a single for these so they can all be enabled/disabled at the same time? For example. `#ifdef ALLOW_PREVIEW_MODE`... What does this preview mode do?

3. Over-use of `auto`: I don't believe using `auto` for everything improves the code except in a very limited number of cases. In most cases it means the reader now has to figure out the type from the value assigned, and could cause the type of the variable to be unintentionally changed if the value is changed.

4. The weird tags in ISavedGame::read and possible ABI weirdness when reading/writing. I'll need to spend a bit more time looking over this so don't worry about this too much unless you see a glaring issue.

5. Build failures from appveyor - although at the moment I don't know what failures these are. One of the other maintainers set this up so I'm not sure where I can see the build details...

shared/qcommon/ojk_saved_game.cpp
bibendovsky: I will in next update.

shared/qcommon/ojk_saved_game.cpp
bibendovsky: I'll remove the first check if there is no other checks will be added.

bibendovsky:
> And 2013 doesn't even support constexpr properly.

I'll fix it the next update.

> It was brought up in discussions that the direct use of the C++ object in the interface won't be guaranteed to work in all cases with mixed library versions or compilers.

I hope at least on same platform but with different compiler abstract classes are ABI compatible.

> Also, isn't the get module API function Extern C?

It is, to disable C++ name mangling.

> Throwing exceptions from ISavedGame is a no go.

But what about, for example, function pointer Error in game_import_t? This pointer initialized in engine with function Com_Error which __throws__ an exception. Here the same story. There is no implementation of ISavedGame interface in game/cgame library only in engine.

> You have a number of if (is_preview_mode_) conditions wrapped in #if 0.

Just leftovers of old code. That was immitation of qbSGReadIsTestOnly for reading saved game in SG_GetSaveGameComment to avoid calling Com_Error in case of inability to read data.

> Over-use of auto

Not a problem to fix in next update.

> The weird tags in ISavedGame::read and possible ABI weirdness when reading/writing.

In the end any template method calls one of the ISavedGame's non-templated one. So I don't see any ABI issues here.

> Build failures from appveyor

This is work in progress branch. I didn't merge it with the latest yet.
Two conflicts for now.

__code/game/g_public.h__

HEAD
#include "qcommon/ojk_i_saved_game_fwd.h"
#define	GAME_API_VERSION	8
= = = = = = =
#define	GAME_API_VERSION	9
master

__code/server/sv_game.cpp__

void SG_ReadCvars(void)
{
HEAD
    int iCount;
    const char* psName;
    const char* psValue;
= = = = = = =
	int		iCount = 0;
	char	*psName = nullptr;
	char	*psValue = nullptr;
master

mrwonko:
I like constexpr and strongly support upgrading to the latest version of VS2015. If it were for me we'd upgrade to C++17 compilers ASAP. Shunning C++ because the original code is in C is silly.

That said, I've not looked at these changes so I can't comment on whether they constitute good C++.

bibendovsky:
> I like constexpr and strongly support upgrading to the latest version of VS2015.

BTW, newer version of Visual Studio may be installed alongside with old one(s).

ensiform:
I'm not speaking in terms for me, I have 2015 and 2013.  Some users don't though

xycaleth:
> I hope at least on same platform but with different compiler abstract classes are ABI compatible.

Passing a *pure virtual* class should be relatively safe. Most major compilers follow the same ABI for this - I've used this assumption in the renderer: https://github.com/JACoders/OpenJK/blob/master/codemp/rd-common/tr_public.h#L343

> But what about, for example, function pointer Error in game_import_t? This pointer initialized in engine with function Com_Error which throws an exception. Here the same story.

Throwing user-defined types is the main problem here, e.g. `ojk::Exception`. `Com_Error` throws an integer. We have an open issue as well for exceptions across library boundaries (#824) but it should be safe to throw basic types. I've tested this on clang and gcc and haven't seen any bug reports yet...

> Just leftovers of old code. That was immitation of qbSGReadIsTestOnly for reading saved game in SG_GetSaveGameComment to avoid calling Com_Error in case of inability to read data.

Okay. Do you think we can get rid of it? If it's not used, then there shouldn't be any reason not to remove it.

> In the end any template method calls one of the ISavedGame's non-templated one. So I don't see any ABI issues here.

I've looked over this again and everything looks more or less fine. I still think it's a bit over-engineered but it works :)

However, I would say that it's a bit misleading to use the name `ISavedGame` when it's not strictly an interface and especially when it's an interface that's shared between the  engine and library code. I think the pure virtual functions should stay in `ISavedGame`, and then split off another class which makes use of this class (and is specific to the library code).

And one last thing, please can you use tabs for indentation instead of spaces? At the moment there's a mix between your code and the existing code!

bibendovsky:
> Throwing user-defined types is the main problem here

Then I'll try to avoid throwing the exception completely. BTW does "const char*" is ok for throw?

> However, I would say that it's a bit misleading to use the name ISavedGame

Agreed, I'll migrate such methods into another class.

> please can you use tabs for indentation instead of spaces
No problem.

bibendovsky:
Update 2.
JO: Fixed saving/loading original saved game
Moved template method of interface into separate class.
Replaced spaces with tabs.
Increased game api version.

code/client/cl_main.cpp
```diff
@@ -1181,6 +1181,8 @@ void CL_InitRef( void ) {
 
 	rit.SV_PointContents = SV_PointContents;
 
+    rit.saved_game = &ojk::SavedGame::get_instance();
```
xycaleth: Tabs instead of spaces :)

code/game/G_Timer.cpp
```diff
@@ -220,8 +237,24 @@ void TIMER_Load( void )
 			assert (sizeof(g_timers[0]->time) == sizeof(time) );//make sure we're reading the same size as we wrote
 
 			//Read the id string and time
-			gi.ReadFromSaveGame( INT_ID('T','M','I','D'), (char *) tempBuffer, 0, NULL );
-			gi.ReadFromSaveGame( INT_ID('T','D','T','A'), (void *) &time, sizeof( time ), NULL );
+			saved_game.read_chunk(
+				INT_ID('T', 'M', 'I', 'D'));
+
+			auto sg_buffer_data = static_cast<const char*>(
```
xycaleth: Remove the auto here please.

code/rd-vanilla/G2_misc.cpp
```diff
@@ -1760,162 +1770,212 @@ void *G2_FindSurface(const model_s *mod, int index, int lod)
...
+		return;
+	}
 
-	// now lets start putting the data we care about into the buffer
-	char *tempBuffer = pGhoul2Data;
 
 	// save out how many ghoul2 models we have
-	*(int *)tempBuffer = ghoul2.size();
-	tempBuffer +=4;
+	auto model_count = ghoul2.size();
```
xycaleth: Remove the auto here please! Cast to int or size_t and then remove all the decltype expressions below. I think it makes things easier to read.

code/rd-vanilla/G2_misc.cpp
```diff
@@ -1760,162 +1770,212 @@ void *G2_FindSurface(const model_s *mod, int index, int lod)
...
 		// no, ok, well, done then.
 		return;
 	}
 
-	// this one isn't a define since I couldn't work out how to figure it out at compile time
-	const int ghoul2BlockSize = (intptr_t)&ghoul2[0].mTransformedVertsArray - (intptr_t)&ghoul2[0].mModelindex;
-
 	// now we have enough instances, lets go through each one and load up the relevant details
-	for (int i=0; i<ghoul2.size(); i++)
+	for (decltype(model_count) i = 0; i < model_count; ++i)
```
xycaleth: Change the decltype here for `int`.

code/rd-vanilla/G2_API.cpp
```diff
@@ -1401,6 +1413,15 @@ qboolean G2API_StopBoneAngles(CGhoul2Info *ghlInfo, const char *boneName)
 	return ret;
 }
 
+#ifdef JK2_MODE
```
xycaleth: What's the benefit of not compiling all of these in JK2? They're not used by JK2 so they won't ever be run.

code/rd-vanilla/G2_misc.cpp
```diff
@@ -1760,162 +1770,212 @@ void *G2_FindSurface(const model_s *mod, int index, int lod)
 	return (void *)current;
 }
 
+#if 1
```
xycaleth: Can we remove the #else block?

code/server/sv_savegame.cpp
```diff
@@ -111,169 +108,30 @@ static const char *GetString_FailedToOpenSaveGame(const char *psFilename, qboole
...
 //	we didn't do this then we'd run out of quake file handles after the 8th load fail...
 //
 void SG_Shutdown()
 {
-	if (fhSaveGame )
-	{
-		FS_FCloseFile( fhSaveGame );
-		fhSaveGame = NULL_FILE;
-	}
+	auto& saved_game = ojk::SavedGame::get_instance();
```
xycaleth: Remove the `auto` here?

code/server/sv_savegame.cpp
```diff
@@ -860,78 +880,152 @@ static char *SG_GetSaveGameMapName(const char *psPathlessBaseName)
...
 {
-	if(!psPathlessBaseName)
+	if (!base_name)
 	{
-		return qfalse;
+		return false;
 	}
 
-	if (!SG_Open(psPathlessBaseName))
+	auto& saved_game = ojk::SavedGame::get_instance();
```
xycaleth: Remove auto

code/server/sv_savegame.cpp
```diff
@@ -1040,7 +1149,9 @@ qboolean SG_WriteSavegame(const char *psPathlessBaseName, qboolean qbAutosave)
 		SG_StoreSaveGameComment(va("--> %s <--",psMapName));
 	}
 
-	if(!SG_Create( "current" ))
+	auto& saved_game = ojk::SavedGame::get_instance();
```
xycaleth: Remove auto

code/server/sv_savegame.cpp
```diff
@@ -1065,536 +1183,154 @@ qboolean SG_WriteSavegame(const char *psPathlessBaseName, qboolean qbAutosave)
...
+			::sv.timeResidual);
+
 		CM_WritePortalState();
 		SG_WriteServerConfigStrings();
 	}
 	ge->WriteLevel(qbAutosave);	// always done now, but ent saver only does player if auto
-	SG_Close();
-	if (gbSGWriteFailed)
+
+	auto is_write_failed = saved_game.is_failed();
```
xycaleth: Remove auto

codeJK2/game/CMakeLists.txt
```diff
@@ -232,7 +232,11 @@ set(JK2SPGameCommonFiles
 	"${SPDir}/qcommon/q_shared.cpp"
 	"${SPDir}/qcommon/q_shared.h"
 	"${SPDir}/qcommon/strippublic.h"
-	
+	"${SPDir}/rd-common/mdx_format.h"
```
xycaleth: The game module shouldn't be including renderer headers apart from tr_public.h. Why is this being included?

codeJK2/game/Q3_Interface.cpp
```diff
@@ -9262,9 +9262,8 @@ void Interface_Init( interface_export_t *pe )
 	pe->I_FreeVariable			=	Q3_FreeVariable;
 
 	//Save / Load functions
-	pe->I_WriteSaveData			=	(int(*)(unsigned int, void *, int))gi.AppendToSaveGame;
-	pe->I_ReadSaveData			=	gi.ReadFromSaveGame;
 	pe->I_LinkEntity			=	ICARUS_LinkEntity;
+    pe->saved_game = gi.saved_game;
```
xycaleth: Tabs/spaces

codeJK2/game/g_public.h
```diff
@@ -25,7 +25,7 @@ along with this program; if not, see <http://www.gnu.org/licenses/>.
 #define __G_PUBLIC_H__
 // g_public.h -- game module information visible to server
 
-#define	GAME_API_VERSION	9				// Bumped up to 8, since we support JA's engine now --eez
+#define	GAME_API_VERSION	10				// Bumped up to 8, since we support JA's engine now --eez
```
xycaleth: Remove the comment here please :) It's slightly out of date now

code/rd-common/tr_public.h
```diff
@@ -112,7 +112,7 @@ typedef struct {
 															const int passEntityNum, const int contentmask,
 															const EG2_Collision eG2TraceType, const int useLod );
 
-	qboolean			(*SG_Append)						( unsigned int chid, const void *pvData, int iLength );
+	ojk::ISavedGame* saved_game;
```
xycaleth: `REF_API_VERSION` needs to be increased for any API changes.

codeJK2/icarus/TaskManager.cpp
```diff
@@ -1638,33 +1639,56 @@ int CTaskManager::SaveCommand( CBlock *block )
...
 		size = bm->GetSize();
-		(m_owner->GetInterface())->I_WriteSaveData( INT_ID('B','S','I','Z'), &size, sizeof( size ) );
+
+		saved_game.write_chunk<int32_t>(
+			INT_ID('B', 'S', 'I', 'Z'),
+			size);
 		
 		//Save out the raw data
-		(m_owner->GetInterface())->I_WriteSaveData( INT_ID('B','M','E','M'), bm->GetData(), size );
+        auto raw_data = static_cast<const uint8_t*>(bm->GetData());
```
xycaleth: Change to tabs and remove auto

codeJK2/icarus/interface.h
```diff
@@ -81,10 +81,11 @@ typedef struct interface_export_s
 
 	//Save / Load functions
 
-	int				(*I_WriteSaveData)( unsigned int chid, void *data, int length );
+	int				(*I_WriteSaveData)( unsigned int chid, const void *data, int length );
 	int				(*I_ReadSaveData)( unsigned int chid, void *address, int length, void **addressptr/* = NULL */);
 	int				(*I_LinkEntity)( int entID, CSequencer *sequencer, CTaskManager *taskManager );
 
+    ojk::ISavedGame* saved_game;
```
xycaleth: Change to tabs

shared/qcommon/ojk_saved_game.cpp
```diff
@@ -0,0 +1,974 @@
...
+	close();
+}
+
+bool SavedGame::open(
+	const std::string& base_file_name)
+{
+	close();
+
+
+	const auto file_path = generate_path(
```
xycaleth: Remove auto

shared/qcommon/ojk_saved_game.cpp
```diff
@@ -0,0 +1,974 @@
...
+
+		error_message_ =
+			"Failed checksum check for chunk " + chunk_id_string + ".";
+
+		return false;
+	}
+
+	// Make sure we didn't encounter any read errors...
+	if (uiLoaded !=
+		sizeof(ulLoadedChid) +
```
xycaleth: Can we move all of this addition into a variable? And then the `#ifdef JK2_MODE` block can be added on without it looking so messy.

xycaleth:
There's still a lot of use of `auto` - I've picked out ones I saw when skimming through but I think there's still a lot more.

shared/qcommon/q_math.h
```diff
@@ -180,6 +184,29 @@ typedef struct cplane_s {
 	byte	type;			// for fast side tests: 0,1,2 = axial, 3 = nonaxial
 	byte	signbits;		// signx + (signy<<1) + (signz<<2), used as lookup during collision
 	byte	pad[2];
+
+
+#ifdef __cplusplus
```
xycaleth: Saving/loading game files shouldn't be in this file. This file is shared by the MP game as well now, and I'd like the shared/ folder to be agnostic of SP or MP.

bibendovsky: Should I fix only this class or all?

xycaleth:
Thanks for sticking with this. New changes look good. I've just noticed one last problem - all the new save game files are in the shared/qcommon folder. This folder is shared by SP and MP, and obviously MP doesn't need it. Can you move these files into code/qcommon? That folder will be used only by the SP code.

bibendovsky:
>Can you move these files into code/qcommon?

No problem.

xycaleth:
@bibendovsky Hey, your changes are causing some issues with our build servers (which we need to link up to GitHub still...). Here's some of the error logs:

https://jk.xd.cm/builders/linux/builds/684/steps/make/logs/stdio

https://jk.xd.cm/builders/linux-64/builds/788/steps/make/logs/stdio

https://jk.xd.cm/builders/linux/builds/685/steps/make/logs/stdio

Please can you create a new PR to fix these problems?

ensiform:
@bibendovsky People are reporting issues with saves and some features in the maps not working now.  Is this because of old saves broken or is this an issue with the code changes more likely?  Seems either entity or ICARUS related.

#889 #902 and  #896 may also be related, not sure.

PR #851, opened 2016-07-18
Merge in "some" of the code?

dusty22:
This is a fake pull request; I was wondering if I could merge in some of the less gameplay-heavy features of my mod, like some of the new cvars.

I just wanted to ask permission before I started making commits on OpenJK.


xycaleth:
You're going to need to be a bit more specific about what you want to add. We're not going to read through 50000 lines of changes :p

ensiform:
This is essentially code from jke which we don't need in the base OpenJK repository.

dusty22:
No, no JKE code.

These are what I was going to add:
g_autoRoll - default is 1, if 1 default JA behavior, if 0 you have to be holding USE to roll
g_char_forcePowerMax - requires level restart, controls number of player force points
g_char_forceRegen* - requires level restart, controls force point regen rate of player
g_char_parryBonus* - same as .sab file setting but applied on savegame load for player
g_char_breakParryBonus* - same as .sab file setting but applied on game load for player
g_handicap - can go up to 200 now
g_saberDamageScale - this scales saber damage, and is saved with your savegame. This effect stacks on top of the effect caused by g_saberMoreRealistic.
g_saberDeflectAutoAim - if 0, blaster deflections are not auto-aimed but go in the direction of the crosshair with randomness
based on Saber Defense level
g_saberForceDrains - if 1 special moves drain the amount of FP from the next cvar, katas not affected
g_saberForceDrainAmount - if g_saberForceDrains is 1, special moves drain this amount of FP, katas not affected
g_saberLockSuperBreaks - if 1 default JA behavior of saber locks, if 0 no super breaks (1-hit KOs) ever happen after saber locks
g_saberMoreRealistic - no longer write-protected (already in OpenJK right?)
g_weaponVelocity, g_weaponAltVelocity - use this to multiply the speed of most projectiles

- visual effect for blocking Force Lightning with your lightsaber
- rolls and acrobatics are possible in 1st Person now

It would be a little surgical. Not just a straight cut and paste on some things, because some are mixed in slightly with other gameplay features.

Razish:
I think I would only want the visual effect for blocking force lightning with saber.

dusty22:
Too much random gameplay stuff to arbitrarily throw into OpenJK? I can see that.

I feel like at the very least though the g_char cvars might be good though. Otherwise I don't think there's any way to save player force point changes. (when you change these things via the playermodel cmd it isn't saved)

dusty22:
After looking through commits, I do see how most of the cvars wouldn't really fit into OpenJK's style.

Another thing I'm working on that might be of interest is that I've added more values to the weapons.dat. But it's not completely finished or tested.

dusty22:
I'll post again when the weapons.dat fields are done.

xycaleth:
I'm closing this PR as I'm just getting emails for every commit you're making to your repository, and this isn't really a PR. If you want to submit something specific, feel free to make another PR!

PR #855, opened 2016-08-01
Autocomplete fixes/hacks when pressing Enter

eternalcodes:
Possible solution for matching jamp autocomplete behavior when pressing Enter?


ensiform:
I don't particularly want to restore enter magic completion.

Also, you'll want to fix your changes in the cl_keys file so that it only applies to the autocomplete stuff and not the extra changes you made to the keys table.

xycaleth:
I think it would be nice to have this feature behind a cvar which defaults to off.

eternalcodes:
I made that change to the keys table because /bind a, /bind k, etc. would autocomplete as something else.

ensiform:
Autocomplete on enter is intended to only function for the first argument, so you may want to introduce a check for the enter used so that it doesn't do argument completion.  But you might run into odd things with qui gon etc.

eternalcodes:
People are used to and prefer short-handing commands. Like xycaleth said, we could put this feature behind a cvar so that we're not forcing people to adapt to using the tab key.

ensiform:
That's correct but quite frankly does nobody understand the concept of adapting to changes at all?  It's not a game change feature therefore you get used to it.

ensiform:
Qui Gon doesn't work because of what I mentioned earlier.

ensiform:
It needs to be cvar toggle imo.  With default no enter key support please.

eezstreet:
good to merge?

Issue #756, opened 2015-11-24
Endianess Functions use platform-dependent int-types


mrwonko:
```
short ShortSwap( short l );
int LongSwap( int l );
```

That should be `int16_t`/`int32_t`. It would also be nicer to use inline functions instead of defines for `LittleShort` etc. The swapping could also be generalized using a template, but that may have to wait until the mp libraries are C++?

This is mostly a note to myself; I'm just too busy to do this right away. (Because I'd like to do it cleanly, and that takes time.)

ensiform:
>This is mostly a note to myself; I'm just too busy to do this right away. (Because I'd like to do it cleanly, and that takes time.)

Wouldn't these make more sense in a mental note document on google or something? =p

mrwonko:
> Wouldn't these make more sense in a mental note document on google or something? =p

I don't see how this is not an issue with OpenJK.

RostakaGmfun:
I'm currently working on this, PR will be opened ASAP.

bmwalters:
This was fixed by #856, no?

xycaleth:
Looks like it!

PR #856, opened 2016-08-01
Refactor byte order utilities

RostakaGmfun:
Mostly relates to #756.

* Reduce code duplication by moving byte order utilities to `shared/` folder
* Rewrite byte order utilities to use compiler builtins where possible for maximum performance.
* Declare byte order utilities as `static inline`.
* Use platform-independent types for byte order functions.
* Remove commented-out code.

I also would like to discuss the `Q_3_BIG_ENDINAN`/`Q3_LITTLE_ENDIAN`, as they seem to be redundant. The `ntohs`/`htons` and friends can be used instead.

P. S. I am new to the codebase.


shared/qcommon/q_platform.h
```diff
@@ -227,11 +227,74 @@ along with this program; if not, see <http://www.gnu.org/licenses/>.
...
 
 // endianness
-void CopyShortSwap( void *dest, void *src );
-void CopyLongSwap( void *dest, void *src );
-short ShortSwap( short l );
-int LongSwap( int l );
-float FloatSwap( const float *f );
+// Use compiler builtins where possible for maximum performance
+#include <stdint.h>
+#if defined(__clang__)
```
xycaleth: Can this block be put in an `#else` block? There may be other compilers which we haven't tested with and there needs to be some fallback.

RostakaGmfun: Done

shared/qcommon/q_platform.h
```diff
@@ -227,11 +227,76 @@ along with this program; if not, see <http://www.gnu.org/licenses/>.
...
 
 // endianness
-void CopyShortSwap( void *dest, void *src );
-void CopyLongSwap( void *dest, void *src );
-short ShortSwap( short l );
-int LongSwap( int l );
-float FloatSwap( const float *f );
+// Use compiler builtins where possible for maximum performance
+#include <stdint.h>
+#if defined(__GNUC__) || defined(__GNUG__)
```
xycaleth: I think this needs to be `#if !defined(__clang__) && (defined(__GNUC__) || defined(__GNUG__))` otherwise for clang it will enter this block, and then not define the functions. Do you agree?

xycaleth:
> I also would like to discuss the Q_3_BIG_ENDINAN/Q3_LITTLE_ENDIAN, as they seem to be redundant. The ntohs/htons and friends can be used instead.

Can you expand on this? Do you mean there's no point having both defines?

And by the way - thanks for submitting this pull request! It's really appreciated :)

shared/qcommon/q_platform.h
RostakaGmfun: I agree, fixed

RostakaGmfun:
@xycaleth Regarding `Q3_BIG_ENDINAN`/`Q3_LITTLE_ENDIAN`, I meant, that they aren't needed at all, and the logic behind them and `LittleShort` etc. functions can be replaced with library functions `ntohl` and others.

However, I might be wrong, since I have no idea about byte order conventions throughout protocol. I would be grateful for some kind of spec/documentation, if it exists :)

ensiform:
The defines are used elsewhere in the codebase too where big endian structures must be dealt with differently like with model loading.

RostakaGmfun:
@ensiform, Got it, no questions anymore.

xycaleth:
Ah, unforeseen consequence of this commit: `__builtin_bswap16` is not available in GCC 4.7, which our Linux build slave uses:

https://jk.xd.cm/builders/linux-64/builds/752/steps/make/logs/stdio

bswap32 is available in 4.7 though: https://gcc.gnu.org/onlinedocs/gcc-4.7.0/gcc/Other-Builtins.html#Other-Builtins

Please can you make another PR to fix this issue? Thanks1 :)

ensiform:
I don't really see the necessary benefits to using these builtin replacements when they are only available on gcc and friends.

Razish:
Probably because `// Use compiler builtins where possible for maximum performance`
Original implementation is still there as a fallback. It's not like gcc is an invalid compiler.

ensiform:
Didn't say it was.

PR #860, opened 2016-08-03
Fix byte order utils for gcc < 4.8

RostakaGmfun:
This should fix the gcc 4.7 build, which was broken by #856.

Tested with gcc 4.7.

BTW, How about enabling CI on PRs for Linux?

### DIFF
--- a/shared/qcommon/q_platform.h
+++ b/shared/qcommon/q_platform.h
@@ -229,13 +229,17 @@ along with this program; if not, see <http://www.gnu.org/licenses/>.
 // endianness
 // Use compiler builtins where possible for maximum performance
 #include <stdint.h>
-#if !defined(__clang__) && (defined(__GNUC__) || defined(__GNUG__))
-
-// gcc
+#if !defined(__clang__) && (defined(__GNUC__) || defined(__GNUG__)) \
+            && ((__GNUC__ * 100 + __GNUC_MINOR__) >= 403)
+// gcc >= 4.3
 
 static inline uint16_t ShortSwap(uint16_t v)
 {
+#if __GNUC_MINOR__ >= 8
     return __builtin_bswap16(v);
+#else
+    return (v << 8) | (v >> 8);
+#endif // gcc >= 4.8
 }
 
 static inline uint32_t LongSwap(uint32_t v)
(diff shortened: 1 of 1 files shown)


shared/qcommon/q_platform.h
```diff
@@ -229,13 +229,17 @@ along with this program; if not, see <http://www.gnu.org/licenses/>.
 // endianness
 // Use compiler builtins where possible for maximum performance
 #include <stdint.h>
-#if !defined(__clang__) && (defined(__GNUC__) || defined(__GNUG__))
-
-// gcc
+#if !defined(__clang__) && (defined(__GNUC__) || defined(__GNUG__)) \
+            && ((__GNUC__ * 100 + __GNUC_MINOR__) >= 403)
```
xycaleth: I'm not too fond of this. Can you check `__GNUC__ >= 4 && __GNUC_MINOR_C >= 3` instead?

RostakaGmfun: Done

shared/qcommon/q_platform.h
```diff
@@ -229,13 +229,17 @@ along with this program; if not, see <http://www.gnu.org/licenses/>.
 // endianness
 // Use compiler builtins where possible for maximum performance
 #include <stdint.h>
-#if !defined(__clang__) && (defined(__GNUC__) || defined(__GNUG__))
-
-// gcc
+#if !defined(__clang__) && (defined(__GNUC__) || defined(__GNUG__)) \
+            && __GNUC__ >=4 && __GNUC_MINOR__ >= 3
```
xycaleth: Heh, sorry - I've just realised this will cause other issues! It will only be true if the version is 4.3, or 5.3, or 6.3, but 5.1, or 5.2 won't be true. I think the original way of doing this was probably better now :)

RostakaGmfun: Hell yeah :smile:

shared/qcommon/q_platform.h
dionrhys: I actually think the original way this was done was the right way. By changing the code to `__GNUC__ >= 4 && __GNUC_MINOR_C >= 3`, the check would fail for versions 5.0-5.2, 6.0-6.2, etc.

xycaleth: Yep, this change was reverted before merging :)

PR #874, opened 2016-09-25
Pick up date from SOURCE_DATE_EPOCH, for reproducible builds

smcv:
The goal of reproducible builds is that a rebuild of the same source
code with the same compiler, libraries, etc. should result in the same
binaries. SOURCE_DATE_EPOCH provides a standard way for build systems
to fill in the date of the latest source change, typically from a git
commit or from metadata like the debian/changelog in Debian packages.

This does not change anything if SOURCE_DATE_EPOCH is not defined;
the intention is that a larger build system like a Debian package
will define it.

Please see https://reproducible-builds.org/ for more information about
reproducible builds.

Similar to https://github.com/ioquake/ioq3/pull/225.

---

If desired you could use this in your buildbot's script by setting the environment variable `SOURCE_DATE_EPOCH` to the number of seconds since 1970-01-01 00:00 UTC (the Unix epoch), for example from `git show --pretty=format:%ct --no-patch` to use the commit timestamp of the latest commit.


code/qcommon/q_shared.h
```diff
@@ -80,6 +80,10 @@ along with this program; if not, see <http://www.gnu.org/licenses/>.
 #define HOMEPATH_NAME_MACOSX HOMEPATH_NAME_WIN
 #endif
 
+#ifndef SOURCE_DATE
```
xycaleth: Can you add this to the `shared/qcommon/` folder instead? Perhaps in `q_platform.h`  if you think it fits there.

I want to try to avoid duplication of code between the `code/` and `codemp/` folders as much as possible.

smcv:
Updated commit moves `SOURCE_DATE` to `q_platform.h` as requested.

ensiform:
I'll let @xycaleth merge this one since the changes requested were his.

code/qcommon/q_shared.h
smcv: > Can you add this to the `shared/qcommon/` folder instead?

Done. Could you re-review, please?

smcv:
@xycaleth, any further comment on this? I think I've done what you asked. I am not aware of a way to clear the "changes requested" flag (I'm surprised it wasn't cleared by me pushing a new version actually).

xycaleth:
I'm not sure what I commented on now, but it all looks good to me now.

smcv:
> I'm not sure what I commented on a few months ago

You didn't like the definition of `SOURCE_DATE` being duplicated in `code{,mp}/qcommon/q_shared.h` and asked me to move it to `q_platform.h`. `q_shared.h` is where I put it in some very similar ioquake3 and iortcw commits, but is less good for OpenJK because `q_shared.h` is duplicated, whereas there is only one `q_platform.h`.

ensiform:
Personally I don't like the change in behavior anyway. It would make more sense to have a way to check the source date.  The build date is preferable for most people. TimeDoctor got confused as to why the date in the builds was way wrong for a while sometime after you merged that in ioq3.

smcv:
> Personally I don't like the change in behavior anyway

Er, unless I got something very wrong, if you don't define `SOURCE_DATE_EPOCH` yourself there is no change in behaviour? That was certainly my intention.

> The build date is preferable for most people

Those people shouldn't define `SOURCE_DATE_EPOCH` then.

smcv:
https://reproducible-builds.org/specs/source-date-epoch/ explains this better than I can.

ensiform:
"These people". Sure, but the default ("official") builds have it defined in ioquake3 which cause(s)/(d) confusion in the past.

smcv:
> the default ("official") builds have it defined in ioquake3

Oh, that's news to me... I didn't expect they'd do that. Presumably it's taken from the commit timestamp, like I mentioned as a possibility in this PR?

My interest in this is only that I want to enable it for the builds in Debian, as a step towards improving confidence that if our autobuilders were compromised and made to produce malicious packages, we'd detect it (a lot of other infrastructure is needed for that, but that isn't something that needs to be done per-package, whereas adding `SOURCE_DATE_EPOCH` support does). Whether to enable it on upstream autobuilders is up to the operators of those autobuilders - it's a policy question, whereas this PR just added the mechanism. If you don't want it enabled on OpenJK's upstream autobuilders then that's absolutely fine, and they can continue with `SOURCE_DATE` expanding to `__DATE__`.

PR #879, opened 2016-10-21
Remove constexpr usage

xycaleth:
VS2013 does not support the `constexpr` keyword so it has been removed.

### DIFF
--- a/code/qcommon/ojk_saved_game_class_archivers.h
+++ b/code/qcommon/ojk_saved_game_class_archivers.h
@@ -19,10 +19,7 @@ template<>
 class SavedGameClassArchiver<cplane_t>
 {
 public:
-	static constexpr bool is_implemented()
-	{
-		return true;
-	}
+	enum { is_implemented = true };
 
 	static void sg_export(
 		SavedGameHelper& saved_game,
--- a/code/qcommon/ojk_saved_game_helper_fwd.h
+++ b/code/qcommon/ojk_saved_game_helper_fwd.h
@@ -290,11 +290,8 @@ template<typename T>
 class SavedGameClassArchiver
 {
 public:
-	static constexpr bool is_implemented()
-	{
-		return false;
-	}
-
+	enum { is_implemented = false };
+	
 	static void sg_export(
 		SavedGameHelper& saved_game,
 		const T& instance)
--- a/code/game/g_savegame.cpp
+++ b/code/game/g_savegame.cpp
@@ -768,10 +768,10 @@ static void copy_retail_gclient_to_current(
 	const RetailGClient& src,
 	gclient_t& dst)
 {
-	constexpr size_t src_pre_size = offsetof(RetailGClient, ps.saber[0]);
-	constexpr size_t src_post_offset = offsetof(RetailGClient, ps.dualSabers);
-	constexpr size_t src_post_size = sizeof(RetailGClient) - src_post_offset;
-	constexpr size_t dst_post_offset = offsetof(gclient_t, ps.dualSabers);
+	const size_t src_pre_size = offsetof(RetailGClient, ps.saber[0]);
+	const size_t src_post_offset = offsetof(RetailGClient, ps.dualSabers);
+	const size_t src_post_size = sizeof(RetailGClient) - src_post_offset;
+	const size_t dst_post_offset = offsetof(gclient_t, ps.dualSabers);
 
 	::memcpy(
 		reinterpret_cast<char*>(&dst),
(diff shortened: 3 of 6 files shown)


xycaleth:
@bibendovsky Would you mind taking a look at my changes to your save code? I didn't realise that `constexpr` isn't supported by Visual Studio 2013 :(

mrwonko:
😭 
constexpr ftw!
How long do we plan to support VS2013? Can't we move on to C++17 once that's supported by gcc, clang and msvc?

xycaleth:
I don't feel like it's a good idea to keep moving the minimum compiler requirements, and they're already higher than I'd like them to be. Add to the fact t hat VS2015 barely supports C++14 so we'll be waiting around a very long time for MSVC to support C++17 :P

Archangel35757:
Yay!  ...glad to hear that VS2013 support is being restored-- and I agree that the compiler version should be locked down.

code/qcommon/ojk_saved_game_class_archivers.h
```diff
@@ -19,10 +19,7 @@ template<>
 class SavedGameClassArchiver<cplane_t>
 {
 public:
-	static constexpr bool is_implemented()
-	{
-		return true;
-	}
+	enum { is_implemented = true };
```
bibendovsky: Valid but looks like a "hack".
Isn't static constant field would be more readable?
(i.e., "static const bool is_implemented = true;")

code/qcommon/ojk_saved_game_helper_fwd.h
```diff
@@ -290,11 +290,8 @@ template<typename T>
 class SavedGameClassArchiver
 {
 public:
-	static constexpr bool is_implemented()
-	{
-		return false;
-	}
-
+	enum { is_implemented = false };
```
bibendovsky: Same as above.
Valid but looks like a "hack".
Isn't static constant field would be more readable?
(i.e., "static const bool is_implemented = true;")

bibendovsky:
Since compiler requirements already locked at C++11, I don't understand why to use VS2013 wich supports less C++11 features then VS2015.

code/qcommon/ojk_saved_game_class_archivers.h
xycaleth: The main reason for doing this is described here:

http://stackoverflow.com/questions/2172647/template-metaprogramming-difference-between-using-enum-hack-and-static-const

xycaleth:
We don't have a hard minimum requirement for C++11 - it's more what is supported by the oldest compilers that we want to support.

bibendovsky:
> We don't have a hard minimum requirement for C++11

CMake passes option to GCC/Clang to use this standard.
Also, [compilation guide](https://github.com/JACoders/OpenJK/wiki/Compilation-guide) suggests to use a latest version of Visual Studio.

It would be nice to mention about minimum language standard, features that should be avoid and minimum Visual Studio version in the [compilation guide](https://github.com/JACoders/OpenJK/wiki/Compilation-guide).

ensiform:
CMake does support those options, but I'm not sure it is very well supporting to test those options for MSVC.

code/qcommon/ojk_saved_game_class_archivers.h
mrwonko: I might have gone for std::true_type instead, but it's functionally identical.

ensiform:
IMHO a compiler macro or cmake check would have been better, no?

bibendovsky:
Newer version of CMake has target_compile_features command to specify required features or to write a test program to check for them.

ensiform:
target_compile_features doesn't support Visual Studio because there isn't a way to detect C++ version since __cplusplus in VS is still 2001 era version I think.

bibendovsky:
From CMake [documentation](https://cmake.org/cmake/help/v3.3/manual/cmake-compile-features.7.html#manual:cmake-compile-features(7)) v3.3.2:

> Supported Compilers
>     MSVC: Microsoft Visual Studio versions 2010 through 2015.

bibendovsky:
I doubt they use __cplusplus to check for features. Probably they are using feature matrix/table for each supported Visual C++ compiler.

ensiform:
Then you bump the cmake requirements to 3.3 which isn't on most Linux Distros that are outdated a bit :)  Linux Mint 18 has 3.5.2 which is plenty but I don't know about older distros.

bibendovsky:
https://cmake.org/cmake/help/v2.8.9/cmake.html#command:try_compile

bibendovsky:
Another way is to write a test program for [try_compile](https://cmake.org/cmake/help/v2.8.9/cmake.html#command:try_compile) command. There is no need to update CMake requirements.

PR #884, opened 2016-10-28
Consistently use "static QINLINE" for inline C code

smcv:
The portable idiom for type-safe macro-like constructs in C is to use
"static inline" where C99 inline is supported, or "static __inline"
on compilers that implement that keyword as a compiler-specific
extension (at least gcc, clang and MSVC do), falling back to just
"static" as a last resort on terrible compilers from the distant past.

Using "static QINLINE" everywhere means there is no point in defining
QINLINE to "static inline" on clang, so stop doing that; QINLINE now
consistently expands to Standard C/C++ inline, or __inline on MSVC,
or to nothing if we don't know how to inline functions on this
compiler.

This silences warnings about redundant qualifiers (static static inline)
for all the functions that were already inline.

There are a couple of uses of non-static QINLINE in C++ code; I've
left those intact, since inline has different (more useful)
semantics in C++, and as far as I'm aware all reasonable C++ compilers
implement it correctly.

---
As requested by @xycaleth on #869.

### DIFF
--- a/codemp/game/NPC_move.c
+++ b/codemp/game/NPC_move.c
@@ -95,7 +95,7 @@ NPC_CheckCombatMove
 -------------------------
 */
 
-QINLINE qboolean NPC_CheckCombatMove( void )
+static QINLINE qboolean NPC_CheckCombatMove( void )
 {
 	//return NPCInfo->combatMove;
 	if ( ( NPCS.NPCInfo->goalEntity && NPCS.NPC->enemy && NPCS.NPCInfo->goalEntity == NPCS.NPC->enemy ) || ( NPCS.NPCInfo->combatMove ) )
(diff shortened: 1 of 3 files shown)


ensiform:
Question though: We seem to have QINLINE defined as commented out inline for Mac builds, I wonder why that doesn't work?

smcv:
> We seem to have QINLINE defined as commented out inline for Mac builds, I wonder why that doesn't work?

My guess would be that it didn't work on an ancient compiler on some now-irrelevant version of Mac OS; but I don't have a non-obsolete Mac to test on, so I'm not touching that bit :-)

ensiform:
I don't have any Mac to test on hence paging those two.

smcv:
I believe modern compilers automatically inline static functions if it helps code size or efficiency anyway, whether they're declared `static inline` or just `static` (and conversely, they might treat a `static inline` function as though it was `static` if that would help code size or efficiency). The main  effect of `static inline` is that compilers won't typically warn about a `static inline` function that is never called, whereas for plain `static` they usually do.

Issue #902, opened 2017-01-04
Compiled OpenJK and elevator in t2_rogue will not go up.


Sirius902:
Reporting a bug? Please make sure you've given the following information - thanks!

**Operating system and version:**
Windows 10

**Is this for single player or multiplayer?**
Single Player

**Description of the bug (and if possible, steps to reproduce the bug):**
Compiled OpenJK source code with the only changes being including the memory header in files that required it for std::uninitialized_copy_n but did not have it included (ie. IcarusImplementation.cpp).  I compiled it in Visual Studio (I've tried both 2017 and 2015) and the elevator at the end of the level t2_rogue will not go up when I stand on it.

**What did you expect to happen instead?**
I expected the elevator to go up, the only way to beat the level, just like it does in the regular game and pre-built versions of OpenJK.

ensiform:
Well what exactly did you change it to?

Sirius902:
I only added `#include<memory>` at the top of each file that was giving an error when I built it because it said std::uninitialized_copy_n is not a member of namespace std.

ensiform:
Ah, seems the author of the savegame changes didn't add that include file to where it was necessary then.

Sirius902:
Yeah. I also went to the level t3_hevil and it seems certain triggers aren't working? When I get near the top of the level there's supposed to be a few enemies that spawn and they don't. Also there's an elevator in that level too and it doesn't work. It doesn't even make the noise when you stand on it.

ensiform:
Looks like some icarus stuff is broken with savegames related to the changes made to make them fully compat across platforms. I started a new map on rogue on Linux64 and the elevator works fine.

@xycaleth

Sirius902:
I went to t2_rogue from the title screen without loading a save file and it worked.

Sirius902:
I started a new game to see what would happen and on yavin2 I can't force push down the wall in the obstacle course.

bibendovsky:
Comparing _data size_ with a _pointer size_ in #881 breaks the loading of ICARUS scripts.

Sirius902:
Yeah, I run a 64 bit operating system so that makes sense.

ensiform:
@smcv see above #881 broke saves.

smcv:
[...]

PR #906, opened 2017-01-21
Make sure to #include <memory> where std::uninitialized_copy_n is used

smcv:
#902 mentions that some compilers want this.


smcv:
Ignore this, @bibendovsky was faster than me.

PR #912, opened 2017-03-04
Fix build for OpenBSD

devnexen:


xycaleth:
Hey David, thanks for the pull request! Can you rename those `_X` variables to just `X` (instead of `__X`)?

Variables names beginning with an underscore followed by a capital letter are reserved for the compiler - same goes for variable names.

code/rd-common/mdx_format.h
```diff
@@ -148,7 +148,7 @@ typedef struct
 	// I'm defining this '<' operator so this struct can be used as an STL <map> key...
 	//
 	#ifdef __cplusplus
-	bool operator < (const mdxaCompQuatBone_t& _X) const {return (memcmp(Comp,_X.Comp,sizeof(Comp))<0);}
+	bool operator < (const mdxaCompQuatBone_t& __X) const {return (memcmp(Comp,__X.Comp,sizeof(Comp))<0);}
```
ensiform: What exactly does BSD not like about these variable declarations?

devnexen: They are just already defined :-)

devnexen:
@xycaleth ok doing it

PR #916, opened 2017-03-12
Missing initialization of ri.GL_ExtensionSupported

badock:
Hello,

I just found that "ri.GL_ExtensionSupported" was not initialized in the multiplayer code, 
causing it to crash during the initialization.

This pull request proposes a fix for this issue.

Best regards,
Jonathan Pastor


badock:
It appears to me that the single player game have already been patched:
https://github.com/badock/OpenJK/blob/badock-patch-1/code/client/cl_main.cpp#L1163

(And I was able to run the game)

xycaleth:
Ah yeah, good spot :) In that case all good.

PR #917, opened 2017-03-12
Fix some compiler warnings

smcv:
Please see individual commits for details.

If any of the individual patches are not desired, the others can be cherry-picked.

### DIFF
--- a/code/cgame/cg_local.h
+++ b/code/cgame/cg_local.h
@@ -665,7 +665,7 @@ const char *CG_ConfigString( int index );
 const char *CG_Argv( int arg );
 
 void QDECL CG_Printf( const char *msg, ... );
-void QDECL CG_Error( const char *msg, ... );
+NORETURN void QDECL CG_Error( const char *msg, ... );
 
 void CG_StartMusic( qboolean bForceStart );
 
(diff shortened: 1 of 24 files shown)


ensiform:
I'd like an extra NORETURN defined specifically for func ptrs to avoid the __attribute__ in regular code.

NORETURN_PTR maybe?

smcv:
> I'd like an extra NORETURN defined specifically for func ptrs to avoid the attribute in regular code.

Sure, I'll add NORETURN_PTR.

smcv:
Branch rebased and updated.

smcv:
I've done as requested in @ensiform's review. Please could someone take another look at this?

Razish:
This looks good to me. Next reviewer can go ahead and merge.

smcv:
@xycaleth: better now?

I initially put `NORETURN_PTR` in the same place as `NORETURN`, but moving them both to `shared` does seem nicer.

For a lot of `qcommon.h` and `q_shared.h` there's a balancing act between sharing as much as possible between MP and SP, and keeping those files similar to the ones in ioquake3 for easier merging; but `NORETURN` is a new thing in OpenJK so that reasoning doesn't matter.

xycaleth:
Much better :)

PR #939, opened 2017-09-06
Avoid undefined behavior to fix 'crashing when using force powers' issue

peter-kien:
on Clang builds

Please don't cast negative integers to enum, as it is undefined behavior.

This commit fixes:
https://jkhub.org/topic/9457-openjkmac-crashing-when-using-force-powers/

### DIFF
--- a/code/game/NPC_reactions.cpp
+++ b/code/game/NPC_reactions.cpp
@@ -375,7 +375,15 @@ void NPC_ChoosePainAnimation( gentity_t *self, gentity_t *other, const vec3_t po
 		{
 			self->painDebounceTime = level.time + 4000;
 		}
-		self->painDebounceTime = level.time + PM_AnimLength( self->client->clientInfo.animFileIndex, (animNumber_t) pain_anim );
+
+		if ( pain_anim >= 0 )
+		{
+			self->painDebounceTime = level.time + PM_AnimLength( self->client->clientInfo.animFileIndex, (animNumber_t) pain_anim );
+		}
+		else
+		{
+			self->painDebounceTime = level.time;
+		}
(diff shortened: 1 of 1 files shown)


ensiform:
Can you apply fixes to JK2 SP gamecode and possibly MP as well?

@xycaleth @Razish @eezstreet thoughts?

xycaleth:
Why is `pain_anim` negative in the first place?

peter-kien:
@xycaleth Cf. line 225 in NPC_reactions.cpp

It's set to -1. (I've confirmed this with a debugger.)

ensiform:
This bug was talked about before and brought up as needing fixing but we didn't know the correct fix without changing behavior.

Razish:
I have taken a look at this. Confirmed same issue with Clang 4.2.1 on Linux/x86_64, enum is treated as unsigned causing OOB.
Setting `self->painDebounceTime = level.time` will indeed replicate the original behaviour (and not crash), but I believe the proper fix would be to handle invalid animation IDs inside `PM_AnimLength` itself as done in MP (also in `BG_AnimLength`).
Reading through `NPC_ChoosePainAnimation`, it looks like the intent is for an animation to potentially not play. We could conditionally check for this case and not attempt to get the length of a non-animation. I still think PM_AnimLength should safely handle the case of non-animations instead of relying on bandaid fixes on each caller.

Note: JOSP+JASP return 0 for invalid anim sets and don't check anim indexes. MP assumes valid anim set and returns -1 for invalid anim indexes...this doesn't appear to be explicitly checked though. For sanity, I will normalise it all to 0 length.

I feel rude closing the PR, but I have now applied the "proper" fix to all projects. Thank you for contributing =]

PR #941, opened 2017-10-15
openbsd build fix

devnexen:


kaashif:
Can you actually build this on OpenBSD with this patch added? This patch is needed, since the build fails to pick up `gl.h` otherwise, but even with it, when I try to build I get the following:

```bash
$ cmake ..
<no errors>
$ make
...
[ 11%] Building CXX object code/CMakeFiles/openjk_sp.amd64.dir/__/shared/sdl/sdl_window.cpp.o
In file included from /home/kaashif/src/misc/OpenJK/shared/sdl/sdl_window.cpp:23:
/usr/local/include/SDL2/SDL_syswm.h:72:10: fatal error: 'X11/Xlib.h' file not found
#include <X11/Xlib.h>
         ^~~~~~~~~~~~
1 error generated.
*** Error 1 in . (code/CMakeFiles/openjk_sp.amd64.dir/build.make:2031 'code/CMakeFiles/openjk_sp.amd64.dir/__/shared/sdl/sdl_window.cpp.o')
*** Error 1 in . (CMakeFiles/Makefile2:146 'code/CMakeFiles/openjk_sp.amd64.dir/all')
*** Error 1 in /home/kaashif/src/misc/OpenJK/build (Makefile:152 'all')
```

I think this is because `/usr/X11R6/include` isn't in the include directories. I'm not an expert on CMake, but I added the following line to a few of the CMakeLists.txt:

```
include_directories(SYSTEM /usr/X11R6/include)
```

And it builds and runs perfectly.

Like I said I don't know CMake so I don't know where I'm "supposed" to add it to the include directories (maybe it should be in a conditional to only do it on OpenBSD or something).

devnexen:
In fact I just add those into CMAKE_C_FLAGS/CMAKE_CXX_FLAGS when I launch cmake

devnexen:
Eventually what would be best => find_package X11 (by head) and adding include_directories (and adding the lib into target_link_libraries) accordingly.

ensiform:
But X11 isn't being included directly in OpenJK. It is being included within SDL2, therefore you should be using system libraries of SDL2, no?

devnexen:
Yes I mean it is just to include reliably X11 headers and linking to X11 libraries wherever is the folder but that s not mandatory you can pass through linker/compiler flags indeed.

devnexen:
Any chance it gets merged ? :-)

Razish:
Nice and simple, merged.

Hopefully people can find this issue if they have issue with X11 includes. Agreed that `find_package` would be best approach.

PR #945, opened 2017-12-07
ChangeTeam not logged in siege

toomuchio:
Announcement suppression was preventing log entries as well, useful information since tracking team switches via ClientUserinfoChanged or spawned is unreliable, if a switch is made mid-round ect...

### DIFF
--- a/codemp/game/g_cmds.c
+++ b/codemp/game/g_cmds.c
@@ -578,6 +578,8 @@ void BroadcastTeamChange( gclient_t *client, int oldTeam )
 {
 	client->ps.fd.forceDoInit = 1; //every time we change teams make sure our force powers are set right
 
+	G_LogPrintf( "ChangeTeam: %i [%s] (%s) \"%s^7\" %s -> %s\n", (int)(client - level.clients), client->sess.IP, client->pers.guid, client->pers.netname, TeamName( oldTeam ), TeamName( client->sess.sessionTeam ) );
+
 	if (level.gametype == GT_SIEGE)
 	{ //don't announce these things in siege
 		return;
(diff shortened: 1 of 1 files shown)


Razish:
Good call, thanks for the PR =]

toomuchio:
No worries, understood.

Would overhauling the logging system be out of the question if I was to commit something like that?

These logs are seriously dated in their format (breaking up rounds with "\n-------------\n" ect) and the lack of a plugin api makes it sometimes the easiest way to interface with the server.

I don't mind committing some time to reworking the logs with more meaningful things and removing unnecessary stuff if it's something that would be accepted.

Not sure if there's a valid reason we're sticking to the old quake 3 log format, or just nobody got to revamping it.

Razish:
There has been talks about it. Issue is there _are_ several programs used to scan the console + logs for information. Changing the format in any useful way would make these programs useless.

Perhaps this should be brought to a community discussion at JKHub. If the decision is made to adopt a new format we can start talking about how it would look like.
The next issue would be standardised adoption across mods and engines.

I think data/event oriented logging (non-informational) should be moved to an external stream ideally. I'd suggest prefixes for discriminating, but then you open the can of worms called "multi-line messages".
Catch us JKCommunity Discord if you like.

toomuchio:
Yeah that was my concern....

Instead of forcing a change like that then, we could always introduce a new log format, that can be turned on or off via a cvar. So it can be used in conjunction with the old format or replace it based on the server operators requirements.

Then perhaps overtime mods would update to support the new format and in a few years the legacy log format could be totally removed?

Razish (dismissed):

ensiform:
I think part of the reason for it is because of the way team switching and such works in siege.

Yes it is not good to avoid the log, but at the same time, but it could lead to log spam.

toomuchio:
That's true, so I guess the best course of action is to add a new log format that can be used in tandem.

UserInfoChanged, results in log spam this shouldn't be half as bad as that. I'd rather have a little spam and the information and none at all to be honest.
I've written a team balancing script and it flies off the handles at the moment due to not being able to track mid round switching, some mods always appear as siege to the server itself while they may be doing something else.

ensiform:
We also discovered that team changing in siege happens elsewhere anyway, which never calls that function. Unless the user or menu evokes the /team cmd

toomuchio:
Damn... Have you guys managed to figure out where it happens? Logging should be added to that as well I can amend this commit, if need be.
Really need that information to process team counts correctly for mods.

Razish:
homer-appear

I'm pretty sure these are the concerning call sites?

- https://github.com/JACoders/OpenJK/blob/c1b347d39b6c069c73818bcd7fedd0da2be08342/codemp/game/g_saga.c#L644-L651
- https://github.com/JACoders/OpenJK/blob/c1b347d39b6c069c73818bcd7fedd0da2be08342/codemp/game/g_cmds.c#L820-L823

PR #951, opened 2018-01-18
Set per-target visibility flags in a way compatible with CMake 2.8.9

### LINKED ISSUE #933: Loading a game after dying cause some thing to stop working.

Fedor Core 26(64-bit), game version: OpenJK 1.0.1.1(64-bit executable) 

Single player, Jedi Night Academy 

1. Save your game(I have tried checkpoints and quicksaves but not manual saves) 
2. Get killed. 
3. Load the save, cutscenes will not play or will freeze, most elevators will not work and pulling switches will have no effect. Enemy AI will still be loaded, I am not sure about bosses. Most other things still work. 
***NOTE:***
When these things stop working, saving your game and restarting the whole game, can make them work. Sometimes you can control your character or a clone of them during a cutscene but other than that they do work as normal. 

Everything functions after loading a save game after death.

dmitrio95:
This pull request is intended to fix #933 as well as its duplicates like #780.

As I have already described in #933, the root of this bug is that symbol visibility flags that had been set to fix the similar issue before were effectively (but apparently unintentionally) disabled by bace05c9b3 because of the use of build target properties that are not supported in CMake 2.8.9. The issue could have been fixed by using CMake 2.8.12 for the project builds and the corresponding update of the declared CMake minimum required version. Still, as I understand from the discussion on #933, such an upgrade is not currently acceptable because of the ensuing issues with the build server.

Considering that, I wrote a small function that resolves this problem in a way compatible with the used CMake 2.8.9 and allows to avoid using [deprecated CMake functionality](https://cmake.org/cmake/help/v3.10/prop_tgt/COMPILE_FLAGS.html) while using the latter CMake versions. This change is proposed in this pull request.

In case CMake version upgrade is done, this change can be reverted.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,6 +317,21 @@ if (NOT ("$ENV{SOURCE_DATE_EPOCH}" STREQUAL ""))
 	set(SharedDefines ${SharedDefines} "SOURCE_DATE=\"${source_date}\"")
 endif()
 
+# ojk_add_compile_options(<target> <options>)
+# Should be used to add compiler options on a per-target
+# basis while cmake_minimum_required is set to version
+# below 2.8.12.
+function(ojk_add_compile_options Target Options)
+    if (ARGN)
+        message(SEND_ERROR "Too many arguments for ojk_add_compile_option")
+    endif()
+
+    if (COMMAND target_compile_options)
+        target_compile_options(${Target} PRIVATE ${Options})
+    else()
+        set_property(TARGET ${Target} APPEND_STRING PROPERTY COMPILE_FLAGS " ${Options}")
+    endif()
+endfunction(ojk_add_compile_options)
 
 
 # Files shared across all projects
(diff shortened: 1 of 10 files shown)


ensiform:
I think we would rather raise the minimum. Not sure to what, probably 3 or 3.2 at this point. What Linux distro still has so old cmake that we care about?

dmitrio95:
Debian 7 has, and it seems to be used on openjk build servers ([1](https://jk.xd.cm/buildslaves/debian-x64), [2](https://jk.xd.cm/buildslaves/debian-x86)). That is the reason why linux builds available on [builds.openjk.org](https://builds.openjk.org) are affected by the mentioned bugs.

Of course, if upgrade of CMake on the build servers is possible, raising the minimum would be much better solution.

ensiform:
Looking for comments by razor and xyc

xycaleth:
I’d be for updating cmake to version 3. I can upgrade cmake on the Linux build servers.

gort818:
Just tested with cmake version 3.5.1 on Linux I can finally play the game without restarting after dying. I am definitely for raising the minimum version.

ensiform:
There's a branch being worked on for cmake improvements

xycaleth:
I think this pull request can be closed now since the minimum CMake version has been increased to 3.1 in #972. Build server has also been updated.

dmitrio95:
Nice to hear that! If so then this pull request is definitely not needed anymore.

PR #956, opened 2018-02-02
[SP] "Give Force" limit changed from 100, uses playermodel's max forc…

PotatoMaster:
…e power now

Previously, the console command "give force" only allowed the amount of force power to go up to 100 and not the actual limit of the current playermodel, because it was set to go until FORCE_POWER_MAX, which is defined as 100. Now, if you are using a playermodel with an increased amount set, that amount will be used as the maximum.


dionrhys:
Thanks for the contribution! 😄

PR #964, opened 2018-03-15
fix zlib includes for internal png

jampio:
Compiles on linux now.

### DIFF
--- a/lib/libpng/CMakeLists.txt
+++ b/lib/libpng/CMakeLists.txt
@@ -5,9 +5,7 @@
 
 # This only has effect in this directory (lib/png).
 include_directories(include/)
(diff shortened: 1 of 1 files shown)


lib/libpng/CMakeLists.txt
```diff
@@ -5,9 +5,7 @@
 
 # This only has effect in this directory (lib/png).
 include_directories(include/)
-if(WIN32)
-	include_directories(../zlib/include/)
-endif()
```
sotpapathe, line 10: This test should probably just be changed to `if(UseInternalZlib)`. It will take care of the `WIN32` case where it is set to ON by default and the Linux case when internal libraries are requested.

PR #965, opened 2018-03-15
improve crossbuild arch detection

jampio:
Correctly sets filename architecture suffix when crossbuilding from x86_64 to x86


ensiform:
Wouldn't a toolchain file suffice?

jampio:
Agreed. 

I'd like to propose a few more adjustments. If we agree on those, I can make a pull request.

- Adjust toolchain **linux-i686** to use **-march=i686**.
- Create toolchain **linux-i386** which uses **-march=i386**.
- Adjust CMakeLists.txt to set filename suffix to use CMAKE_SYSTEM_PROCESSOR triplet. This would cause the current i686 toolchain to emit files such as jampgamei686.so

ensiform:
The game looks for 386 doesn't it though?

ensiform:
If we weren't going for compatibility I would rather drop both and just use x86

jampio:
Okay, to not harm compatibility, I've omitted the change to CMakeLists.txt.
As it currently stands, i386 is generically used for any 32-bit platform on linux.

PR #966, opened 2018-03-20
old SDK mod null saber fix

erfg12:


xycaleth:
No need to close I don't think. Unless @erfg12 doesn't want to change his pull request to check it properly.

erfg12:
Never messed with this stuff. Just reverted the code for a quick fix.

ensiform:
Well the changes would be in sv_gameapi.cpp not in the ghoul files afaik

PR #967, opened 2018-03-22
Fix Ghoul2 API on server and UI for legacy mods

ensiform:
Also code cleanup for CGame Ghoul2 API in legacy processing.

### DIFF
--- a/codemp/client/cl_uiapi.cpp
+++ b/codemp/client/cl_uiapi.cpp
@@ -290,6 +290,7 @@ static qboolean CL_G2API_GetBoltMatrix_NoRecNoRot( void *ghoul2, const int model
 		return qfalse;
 	}
 
+	re->G2API_BoltMatrixReconstruction( qfalse );
 	re->G2API_BoltMatrixSPMethod( qtrue );
 	return re->G2API_GetBoltMatrix( *((CGhoul2Info_v *)ghoul2), modelIndex, boltIndex, matrix, angles, position, frameNum, modelList, scale );
 }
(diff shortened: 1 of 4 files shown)


ensiform:
Needs testing.

erfg12:
Compiled, ran with clanmod v1.3.0beta, spawned 10 reelo bots, nothing crashed.

ensiform:
@erfg12 Added new commit that may help with that though I am not sure.

codemp/client/cl_uiapi.cpp
```diff
@@ -290,6 +290,7 @@ static qboolean CL_G2API_GetBoltMatrix_NoRecNoRot( void *ghoul2, const int model
 		return qfalse;
 	}
 
+	re->G2API_BoltMatrixReconstruction( qfalse );
```
xycaleth: I don't think this should be here... I think it might only in cgame because of... https://github.com/JACoders/OpenJK/blob/master/codemp/cgame/cg_players.c#L9252-L9258

codemp/rd-dedicated/G2_API.cpp
```diff
@@ -649,7 +642,7 @@ int G2API_InitGhoul2Model(CGhoul2Info_v **ghoul2Ptr, const char *fileName, int m
 
 qboolean G2API_SetLodBias(CGhoul2Info *ghlInfo, int lodBias)
 {
-	if (ghlInfo)
+	if (G2_SetupModelPointers(ghlInfo))
```
xycaleth: Why are these changed?

ensiform: To be in sync with the original code and rd-vanilla.

codemp/client/cl_uiapi.cpp
ensiform: It is how the original SDK behaves I believe. I was just trying to make it all in sync with original behavior.

ensiform: Please check with that

PR #979, opened 2018-06-13
List of prohibition of sharing

Stalinsky1337RUSUKR:
hHello ,

Today i open the list of prohibition addresses 

currently use on 3 UKR/RUS FFA server

This addresses are not allowed by players due to policy violations. We open so all server use today 🎉 😄

### DIFF
--- a/codemp/server/sv_client.cpp
+++ b/codemp/server/sv_client.cpp
@@ -116,6 +116,6153 @@ static qboolean SV_IsBanned( netadr_t *from, qboolean isexception )
 			return qfalse;
 	}
 
+	const char* адрес = NET_AdrToString( *from );
+
+	if ( !Q_stricmp( адрес, "111.0.244.177" ) ) {
+		return qtrue;
+	}
+	if ( !Q_stricmp( адрес, "55.142.23.148" ) ) {
+		return qtrue;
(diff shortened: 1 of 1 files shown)


ensiform:
Umm there are no actual changes and I don't know what it is you are offering.  IP banning can be gotten around with VPNs.

ensiform:
Just create a ban file that can be shared, that would be better option.

Stalinsky1337RUSUKR:
he changed, but maybe it's too big? i see change and works on server

ban file is limit, but code is not limit

also we have anti vpn, i share later 😉 

i keep list update all months!please accept

Stalinsky1337RUSUKR:
@ensiform

Stalinsky1337RUSUKR:
@Razish

Stalinsky1337RUSUKR:
@xycaleth

Stalinsky1337RUSUKR:
@eezstreet

ensiform:
Your code is in russian and it has caused the entire file to be "changed"  The line changes do not appear because of it.  I cannot accept this anyway, because as I said you can release a banlist file that uses the internal engine banlist functionality with all of the IPs that is not limited like the old addip method.

Issue #985, opened 2018-08-06
JK2 fails to build with gcc 8: std::map with wrongly-typed std::allocator


smcv:
**Operating system and version:**

Debian unstable (development rolling release branch) with gcc-8 8.2.0-3

**Is this for single player or multiplayer?**

JK2 single player

**Description of the bug (and if possible, steps to reproduce the bug):**

Compile with JK2 enabled. Compilation fails with several errors similar to this:

```
In file included from /usr/include/c++/8/map:61,
                 from /home/smcv/src/openjk/codeJK2/game/../game/../../code/ui/../qcommon/../game/ghoul2_shared.h:28,
                 from /home/smcv/src/openjk/codeJK2/game/../game/../../code/ui/../qcommon/q_shared.h:512,
                 from /home/smcv/src/openjk/codeJK2/game/../game/../../code/ui/gameinfo.h:27,
                 from /home/smcv/src/openjk/codeJK2/game/../game/g_local.h:32,
                 from /home/smcv/src/openjk/codeJK2/game/g_headers.h:28,
                 from /home/smcv/src/openjk/codeJK2/game/g_ICARUS.cpp:25:
/usr/include/c++/8/bits/stl_map.h: In instantiation of ‘class std::map<std::__cxx11::basic_string<char>, pscript_s*, std::less<std::__cxx11::basic_string<char> >, std::allocator<pscript_s*> >’:
/home/smcv/src/openjk/codeJK2/game/g_ICARUS.cpp:34:15:   required from here
/usr/include/c++/8/bits/stl_map.h:122:21: error: static assertion failed: std::map must have the same value_type as its allocator
       static_assert(is_same<typename _Alloc::value_type, value_type>::value,
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/8/bits/stl_map.h: In instantiation of ‘class std::map<std::__cxx11::basic_string<char>, int, std::less<std::__cxx11::basic_string<char> >, std::allocator<int> >’:
/home/smcv/src/openjk/codeJK2/game/g_ICARUS.cpp:35:13:   required from here
/usr/include/c++/8/bits/stl_map.h:122:21: error: static assertion failed: std::map must have the same value_type as its allocator
```

**What did you expect to happen instead?**

Successful compilation

**Diagnosis**

[...]

PR #986, opened 2018-08-06
JK2: Remove mismatched allocator from std::map typedefs

smcv:
The Allocator is meant to be one that is suitable for allocating
std::pair<key, value> instances, not value instances, and the
Standard C++ library provided by gcc 8 asserts that the types match.
The closest valid allocator here would be
std::allocator<std::pair<key, value>>, but that is the default for
this template anyway, so we can just omit it.

Closes: https://github.com/JACoders/OpenJK/issues/985  
Bug-Debian: https://bugs.debian.org/905477  
Signed-off-by: Simon McVittie <smcv@debian.org>


smcv:
If this branch is OK, please could someone merge it?

PR #991, opened 2018-10-31
Removed superfluous code to determine the number of compressed bones.

AJSchat:
The compressed bone pool of a Ghoul II animation file (MDXA, .gla) is always
stored at the end of the file. It is unnecessary to iterate through all frames
to determine the number of compressed bones present on big-endian machines.

This change is applied to the vanilla renderers of the single- and multiplayer
code of JK:JA.

### DIFF
--- a/code/rd-common/mdx_format.h
+++ b/code/rd-common/mdx_format.h
@@ -432,10 +432,11 @@ typedef struct {
 	//
 	// (iFrameNum * mdxaHeader_t->numBones * 3) + (iBoneNum * 3)
 	//
-	//  then read the int at that location and AND it with 0x00FFFFFF. I use the struct below simply for easy searches
+	// Then convert the three byte int at that location.
+	// This struct is used for easy searches.
 	typedef struct
 	{
-		int iIndex;	// this struct for pointing purposes, need to and with 0x00FFFFFF to be meaningful
+		byte    iIndex[3];
 	} mdxaIndex_t;
 	//
 	// (note that there's then an alignement-pad here to get the next struct back onto 32-bit alignement)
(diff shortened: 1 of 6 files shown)


xycaleth:
Thanks for the PR @AJSchat.

My main concern with your change is if anybody wants to change the Ghoul2 format in their fork then they will need to know to change this code too - not doing so would cause the wrong number of bones to be calculated. So, it's not safe to assume that the bone pool will always reside at the end of the file.

xycaleth:
@Archangel35757 this isn't the place to advertise your mod...

> Thanks for your prompt reply xycaleth!

No problem :)

> I have just created another commit which only applies the new code to version six of the MDXA format. For any other version it falls back to the original behavior attempting to find the maximum bone index by iterating through all bones across frames. It should be safe to assume that anyone wishing to extend the Ghoul II MDXA file format will alter the internal version number too, right? Or do you have a different opinion regarding the matter?

I don't feel like the change here is necessary. The code for the change is fine, but why complicate the code for an insignificant performance improvement? + it's kind of weird to me to check for a specific version when loading.

As far as I know, the code here isn't a bottleneck and `.gla` files are very rarely loaded, so I don't think any change here is needed.

> I would also like to discuss altering the mdxaIndex_t structure. [...] Converting the three bytes to an integer directly is more clear, faster on big-endian machines and is endian independent.

I'm all for making the code clearer to read so this sounds good to me 👍

AJSchat:
I can see where you're coming from. The change is indeed unnecessary when taking future expandability of the Ghoul II format in account and becomes even messier checking for versions of the format. I suppose the change would make more sense if OpenJK's goal wasn't to provide a clean base for other mods to build on.

I will revert the changes made to the MDXA loading functions and commit the other small change to the format in the coming days.

AJSchat:
I just reverted the change and altered the mdxaIndex_t struct as discussed earlier.

Please let me know if any other change is required here or if the PR is approved or rejected as-is. Thank you in advance.

PR #993, opened 2018-11-29
Add Docker support for dedicated server

fcrespel:
Hello,

This PR adds the possibility to build a Docker image for OpenJK Linux-based dedicated servers. Docker makes it much easier to run any program in an isolated "container" (see [Wikipedia](https://en.wikipedia.org/wiki/Docker_(software))).

The Dockerfile uses [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) to compile openjkded and jampgame for both i386 and x86_64, leaving the choice of which to run to the user.

The final image provides the following directory tree and files:

    /opt/openjk
    /opt/openjk/cdpath
    /opt/openjk/cdpath/base
    /opt/openjk/cdpath/base/server.cfg
    /opt/openjk/cdpath/base/jampgamex86_64.so
    /opt/openjk/cdpath/base/jampgamei386.so
    /opt/openjk/homepath
    /opt/openjk/basepath
    /opt/openjk/functions.sh
    /opt/openjk/run.sh
    /opt/openjk/healthcheck.sh
    /opt/openjk/openjkded.i386
    /opt/openjk/openjkded.x86_64

The `homepath`, `basepath` and `cdpath` directories correspond to the `fs_homepath`, `fs_basepath` and `fs_cdpath` Cvars, which are looked up in this order by JKA. Thus, the `cdpath` contains files baked in the Docker image, while the `basepath` should be bound to a volume containing shared assets (assets1.pk3, etc.), and the `homepath` to a volume holding server-specific files (e.g. custom assets, logs, etc.)

A [healthcheck](https://docs.docker.com/engine/reference/builder/#healthcheck) script is included, that will attempt to `getinfo` via UDP every 10s for up to 6 times before marking the container as unhealthy. This can help detecting and restarting crashed servers when doing JKA hosting.

Additionally, the following environment variables are available:

| Variable | Default value | Description |
|----------|------------------|--------------------------------------------|
| OJK_MOD | base | Custom fs_game to load (mod directory) |
| OJK_ARCH | i386 | Server architecture, either i386 or x86_64 |
| OJK_OPTS | +exec server.cfg | Server init options |

To build the image, execute in the source directory:

    docker build -t openjkded:latest .

Supposing you have assetsX.pk3 files in `/home/jka/_shared_/base` and an empty dir in `/home/jka/jka01` for your server files, you may run the server with:

    docker run -d --name jka01 -p 29070:29070/udp -v /home/jka/_shared_:/opt/openjk/basepath -v /home/jka/jka01:/opt/openjk/homepath openjkded:latest

To run a different mod and add custom options:

    docker run -d --name jka01 -p 29070:29070/udp -v /home/jka/_shared_:/opt/openjk/basepath -v /home/jka/jka01:/opt/openjk/homepath -e OJK_MOD=rpmod -e OJK_OPTS="+set fs_basegame JEDI +exec server.cfg" openjkded:latest

To see logs:

    docker logs -f jka01

Note also that automated builds for this Docker image could be setup at [hub.docker.com](https://hub.docker.com), as I have done for the JEDI clan: https://hub.docker.com/r/jediholo/openjkded/ (but "official" ones would be better).

I have been using this image as a base to build an RPMod one for JEDI, and have been running it in production for a few weeks with success. So I felt like contributing it back to OpenJK ;-)

Feel free to ask any question or request adjustments as you see fit.
\- Fabien, a.k.a. Soh Raun.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,10 +4,8 @@ FROM ubuntu:18.04 as builder
 # Install build tools and libraries
 RUN dpkg --add-architecture i386 &&\
 	apt-get -q update &&\
-	DEBIAN_FRONTEND="noninteractive" apt-get -q upgrade -y -o Dpkg::Options::="--force-confnew" --no-install-recommends &&\
 	DEBIAN_FRONTEND="noninteractive" apt-get -q install -y -o Dpkg::Options::="--force-confnew" --no-install-recommends build-essential gcc-multilib g++-multilib cmake libjpeg-dev libjpeg-dev:i386 libpng-dev libpng-dev:i386 zlib1g-dev zlib1g-dev:i386 &&\
-	apt-get -q autoremove &&\
-	apt-get -q clean -y && rm -rf /var/lib/apt/lists/* && rm -f /var/cache/apt/*.bin
+	rm -rf /var/lib/apt/lists/*
 
 # Copy sources
 COPY . /usr/src/openjk
@@ -40,10 +38,8 @@ FROM ubuntu:18.04
 # Install utilities and libraries
 RUN dpkg --add-architecture i386 &&\
 	apt-get -q update &&\
-	DEBIAN_FRONTEND="noninteractive" apt-get -q upgrade -y -o Dpkg::Options::="--force-confnew" --no-install-recommends &&\
 	DEBIAN_FRONTEND="noninteractive" apt-get -q install -y -o Dpkg::Options::="--force-confnew" --no-install-recommends socat libstdc++6 libstdc++6:i386 zlib1g zlib1g:i386 &&\
-	apt-get -q autoremove &&\
-	apt-get -q clean -y && rm -rf /var/lib/apt/lists/* && rm -f /var/cache/apt/*.bin
+	rm -rf /var/lib/apt/lists/*
 
 # Copy binaries and scripts
 RUN mkdir -p /opt/openjk/cdpath/base /opt/openjk/basepath /opt/openjk/homepath
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,13 @@
 # Builder image
-FROM centos:7 as builder
-
-# Install EPEL repository for cmake3
-RUN yum -y install epel-release &&\
-	yum -y clean all
+FROM ubuntu:18.04 as builder
 
 # Install build tools and libraries
-RUN yum -y install gcc gcc-c++ make cmake3 \
-        glibc-devel libstdc++-devel libjpeg-turbo-devel libpng-devel zlib-devel \
-        glibc-devel.i686 libstdc++-devel.i686 libjpeg-turbo-devel.i686 libpng-devel.i686 zlib-devel.i686 &&\
-	yum -y clean all
+RUN dpkg --add-architecture i386 &&\
+	apt-get -q update &&\
+	DEBIAN_FRONTEND="noninteractive" apt-get -q upgrade -y -o Dpkg::Options::="--force-confnew" --no-install-recommends &&\
+	DEBIAN_FRONTEND="noninteractive" apt-get -q install -y -o Dpkg::Options::="--force-confnew" --no-install-recommends build-essential gcc-multilib g++-multilib cmake libjpeg-dev libjpeg-dev:i386 libpng-dev libpng-dev:i386 zlib1g-dev zlib1g-dev:i386 &&\
+	apt-get -q autoremove &&\
+	apt-get -q clean -y && rm -rf /var/lib/apt/lists/* && rm -f /var/cache/apt/*.bin
 
 # Copy sources
 COPY . /usr/src/openjk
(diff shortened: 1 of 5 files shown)


ensiform:
Does "base" specification ignore +set fs_game all together (as it should?)

fcrespel:
> Does "base" specification ignore +set fs_game all together (as it should?)

It didn't, but I've just fixed that 👍

fcrespel:
@ensiform So, what do you think of this PR? could it be merged, and the Docker Hub build setup?

Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+        glibc-devel.i686 libstdc++-devel.i686 libjpeg-turbo-devel.i686 libpng-devel.i686 zlib-devel.i686 &&\
+	yum -y clean all
+
+# Copy sources
+COPY . /usr/src/openjk
+
+# Build i386 arch
+RUN mkdir /usr/src/openjk/build.i386 &&\
+	cd /usr/src/openjk/build.i386 &&\
+	cmake3 -DCMAKE_TOOLCHAIN_FILE=CMakeModules/Toolchains/linux-i686.cmake \
```
madebr: Maybe pass [CMAKE_INSTALL_PREFIX](https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_PREFIX.html) and run `make install`?
This allows the server image to only copy the install tree.

dionrhys:
@ensiform I can take this PR if it's good. I'm not all up to speed on Docker yet, but I should be able to review this in the coming weeks.

dionrhys:
I'm still looking at this in my spare time.

One of the gripes I have at the moment is the sheer size of the CentOS base image:
Docker images list comparing the size of various OS base images

This, combined with all the runtime packages installed on top, results in a total image size of 285MB on my end. (Whereas all the built OpenJK binaries themselves are only 6.87 MBs)

I'm considering adopting the Ubuntu image (since we know we have good support for that already, and they seem to be putting good effort into minimizing their image footprint). But it'd be fab if we could provide an Alpine image too, albeit I haven't investigated this to see how easy it would be to bundle in the required runtime packages (for running OpenJK and for maintaining it remotely).

-----

I'll likely also change some of the output file locations around (OpenJK's game binaries should go into `OpenJK` for example instead of `base`), since we support users running the basejka game and assets from the `base` folder in regular deployments.

-----

This is some very good work so far though! :)

mrwonko:
When it comes to Docker images, my favourites are "scratch" images without anything except your application. This requires you to bundle all runtimes though, which isn't as simple here as it would be with, say, Go.

Almost as light an option are Google's "distroless" images, which I haven't personally tried out but souns like a good choice for you.

fcrespel:
@dionrhys I admit I didn't consider image size when using CentOS 7, it's just the server distro I'm most comfortable with. So, I just tried and managed to build another one from Ubuntu 18.04 (published as "jediholo/openjkded:ubuntu1804"), with a resulting size of 108MB.

If this is the way to go, would you like me to add it to this pull request?

As for Alpine, I don't have experience building images based on it so I'm not sure e.g. how to pack both 64-bit and 32-bit versions and retain mod compatibility... not sure it's worth the trouble?

ensiform:
You would never have mod compatibility with 64bit anyway. There are no mods that are 64bit except new ones that might have released as such since 2013 but I don't know of any specifically.

mrwonko:
You'll need one Docker image per architecture, you can't have both 32 and 64 but in one, to my knowledge.

dionrhys:
@fcrespel I do think basing it on `ubuntu:18.04` would be a good idea then. I also think that it's good that we just use the default amd64 architecture for the OS, but build both i386 and amd64 versions of OpenJK (as seems to work well in the testing I've done!).

If we want to experiment with an Alpine base image, or any other base image, we can do so any time in the future :)

fcrespel:
@dionrhys Alright, so I just pushed the Dockerfile based on `ubuntu:18.04`, which includes both the 32-bit and 64-bit openjkded and jampgame, as well as some suggested changes (OpenJK mod dir, CMAKE_INSTALL_PREFIX + make install). The startup script defaults to the 32-bit version but an environment variable allows switching architectures within the same image.

Besides, I'm not sure there is really a use case for pure 32-bit images? As for Alpine, what I meant with mod compatibility was more related to the C library used by Alpine, which AFAIK is not the usual glibc... but I don't know the implications.

Dockerfile
```diff
@@ -0,0 +1,62 @@
+# Builder image
+FROM ubuntu:18.04 as builder
+
+# Install build tools and libraries
+RUN dpkg --add-architecture i386 &&\
```
dionrhys: The Docker docs recommend against running `apt-get upgrade` or `apt-get dist-upgrade`, "as many of the 'essential' packages from the parent images cannot upgrade inside an unprivileged container". https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

They also state in this section that "Official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required".

So I'm hoping these lines can forego the `upgrade`, `autoremove`, and `clean` bits to make the script a bit less verbose. (And then only need to add i386 arch, update package list, and install the needed packages)

This should be the same for the apt-get invocations further in the file.

fcrespel: Thanks for the feedback, I didn't know about these Debian/Ubuntu best practices :-)
I just commited/tested/pushed the changes.

dionrhys:
Thank you very much for the contribution here! This will be very welcome for server owners, especially if they run quite a few instances of OpenJK. :)

I will set up publishing to Docker Hub, soon™.

PR #994, opened 2018-12-17
Steam Integration.

eezstreet:
Only works on Windows. If com_steamIntegration is 1 on startup (maybe it can default to 1?), the game will look for a steam_api.dll and steam_appid.txt and attempt to initialize the Steam API. With an extra download of these two files, one can have the Steam Overlay and hours tracking without launching OpenJK through Steam itself.

### DIFF
--- a/shared/sys/sys_win32.cpp
+++ b/shared/sys/sys_win32.cpp
@@ -20,6 +20,7 @@ along with this program; if not, see <http://www.gnu.org/licenses/>.
 */
 
 #include "sys_local.h"
+#include "sys_loadlib.h"
 #include <direct.h>
 #include <io.h>
 #include <shlobj.h>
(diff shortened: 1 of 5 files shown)


shared/sys/sys_win32.cpp
```diff
@@ -590,3 +591,87 @@ void Sys_Sleep( int msec )
...
+	if (!gp_steamLibrary)
+	{
+		Com_Printf("Skipping Steam integration shutdown...\n");
+		return;
+	}
+
+	SteamAPI_Shutdown();
+	Sys_UnloadLibrary(gp_steamLibrary);
+	gp_steamLibrary = nullptr;
+}
```
grim-fendango: Missing new line at EOF. Not going to cause any problems but it's best practices to include this

PR #997, opened 2019-01-25
Fixed dismemberment probability bug in SP.

eezstreet:
Fixed a bug in SP (in the original game?) where dismemberment was harder than it needed to be. Also removed cheat flag and added archive flag to g_saberRealisticCombat, debug_subdivision, g_dismemberProbabilities


ensiform:
Good to go despite what github is saying.

PR #1009, opened 2019-04-18
Fix release build with clang 6 compiler

pkubaj:
Don't use variables defined only for debug in release mode.


xycaleth:
What compiler was this breaking on? Asserts should be no-ops with none of the arguments being parsed

pkubaj:
Clang 6 on FreeBSD 11.2.

xycaleth:
According to cppreference, `assert` should be evaluated to nothing by the preprocessor when `NDEBUG` is defined which I would have thought is defined by the implementation even though it doesn't need to.

https://en.cppreference.com/w/cpp/error/assert

@pkubaj Can you check if `NDEBUG` is being defined when you do a release build? I think the better fix here is to define `NDEBUG` ourselves in the `CMakeLists.txt` file if you find it is defined.

pkubaj:
You're right, adding -DNDEBUG fixes build.

PR #1011, opened 2019-05-03
Avoid double promotion via math functions

ShFil119:
it *should* help with vectorization. ;)

### DIFF
--- a/codemp/cgame/cg_draw.c
+++ b/codemp/cgame/cg_draw.c
@@ -256,7 +256,7 @@ static void CG_DrawZoomMask( void )
 
 		CG_DrawPic( 212, 367, 200, 28, cgs.media.binocularOverlay );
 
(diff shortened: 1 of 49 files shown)


ensiform:
Double promotion functions are more well supported in older versions of VS and ensure that there isn't going to be sneaky unexpected precision changes from vanilla whenever possible though.

ShFil119:
So you prefer moving to c++ (`std::` overloaded for each type) or avoiding this changes?

ensiform:
Certainly not. The mp sdk is C

ShFil119:
I've seen issue about performance on rpi 3. (probably it was on forum) 
Comparing to gaining performance via diving tasks into threads, which requires significant changes in engine. This is almost *free*.

//btw I've seen large number of double consts, which can also cause promotion.

PR #1012, opened 2019-05-04
Using proper functions for type at hand.

devnexen:


xycaleth:
Please can you explain what this pull request is for?

devnexen:
Sure. It is to use proper function whether it is float or integer type. Caught compiler warnings while building.

xycaleth:
Thanks for your contribution, @devnexen!

PR #1015, opened 2019-05-14
High-res executable icons for Windows.

SirYodaJedi:
I wasn't sure why the Windows icons were only 32x32 when the source image is 1024x1024. 1024x1024 *is* overkill for Windows, so feel free to cut it down to 256x256.


xycaleth:
Thanks for the PR, @SirYodaJedi

SirYodaJedi:
> Side note: I'm surprised by how many copies of the icon we have. We should consolidate them all at some point 😃
I figured it was due to how Linux distros usually handle icons for `.desktop` links. Some distros, like Ubuntu, support icons in Windows `.ico` format, most apps tend to provide their icons as a set of PNGs at various common resolutions or as a single SVG file. Since the OpenJK icon appears to be created as a bitmap in PhotoShop, having the icons as a set of PNGs probably made the most sense.

ensiform:
We also should really have different versions for mp/SP/jk2 and mp ded probably

smcv:
> We also should really have different versions for mp/SP/jk2 and mp ded probably

#630.

PR #1018, opened 2019-06-25
Improved OOB packet rate limiter

aufau:
This is an redesigned server rate limiter for incoming Out Of Bounds (Connectionless) packets.

The purpose of these patches is to mitigate spoofed IP OOB DOS attacks. This is achieved by two changes:

1. Fixing issue with original q3a rate limiter that was causing very high CPU usage when under such attack, up to the point of rendering server unusable.
2. Allowing "whitelisted" players to fully access OOB server protocol when server is under such attack.

There are 3 new cvars:
- `sv_autoWhitelist` – Enable automatically adding connected players to whitelist
- `sv_maxOOBRate` – Max. incoming OOB packet rate per second. 1000 is default, raising it even more would require code changes.
- `sv_maxOOBRateIP` – Max. incoming OOB packet rate per second for a single IP. 1 is default but there is a burst of 10 times the cvar value and it appears to be enough.

Known issues:
1. Lack of whitelist.dat file access synchronization. Rudimentary version of this was implemented in jk2mv.
2. Concerns have been raised about storing IP in the whitelist.dat file being against the GDPR or other privacy laws. If this is a problem, there is a cvar `sv_autoWhitelist`, or IP could be stored as its digest.

I can explain the attacks addressed by this PR and how they are mitigated in depth, on a private communication channel.

These patches have been tested on actual servers by @Yberion and they managed to mitigate few real attacks that would have crippled the server without them. Also a slightly different version has been part of JK2MV 1.4.1 for over a year.

### DIFF
--- a/codemp/qcommon/files.cpp
+++ b/codemp/qcommon/files.cpp
@@ -818,6 +818,46 @@ fileHandle_t FS_SV_FOpenFileWrite( const char *filename ) {
 
 /*
 ===========
+FS_SV_FOpenFileAppend
+
+===========
+*/
+fileHandle_t FS_SV_FOpenFileAppend( const char *filename ) {
+	char			*ospath;
+	fileHandle_t	f;
+
+	FS_AssertInitialised();
+
+	f = FS_HandleForFile();
+	fsh[f].zipFile = qfalse;
+
+	Q_strncpyz( fsh[f].name, filename, sizeof( fsh[f].name ) );
+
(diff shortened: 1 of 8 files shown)


Yberion:
What about merging this ?

bartrpe:
up, it looks megagood

bartrpe:
Buump? Merge? 
@Razish @ensiform @xycaleth

PR #1019, opened 2019-06-29
Set max line width to 80 characters

0nasi101:
Limit code to 80 columns for compatibility with punch cards and 1970 terminals


xycaleth:
Is this really necessary? 🤔

ensiform:
@xycaleth I'm going to have to agree. I don't see the point.

dionrhys:
I think a bit more work is needed here to integrate compatibility with punch cards and 1970 terminals.

While this is slightly outside the scope of the OpenJK project, I heartily support this endeavour in your own fork of the project.

We'll be delighted to hear from you once you have the project running on a machine like the IBM 2540. Let us know how you get on, by letter or fax :)

Issue #1021, opened 2019-08-27
Segfault in Single Player


Fallstar:
Reporting a bug? Please make sure you've given the following information - thanks!

**Operating system and version:**
Arch Linux, using Wayland

**Is this for single player or multiplayer?**
Single Player

**Description of the bug (and if possible, steps to reproduce the bug):**
Segfault when starting a new game, after character creation.
Multiplayer works fine though, so I doubt it's related to my graphical environment configuration.
I built it today using the AUR pkgbuild.
It happens when using both x11 or Wayland driver for SDL.

**What did you expect to happen instead?**
Start playing.

PS: I don't know if it's relevant, but I get this exact same behavior with Jedi Outcast.

Fallstar:
Nothing explicit, it happens during Game Initialization. Here's the full output: https://hastebin.com/gadiyimuye

Fallstar:
If you don't know what could cause this, I'll build it again with the debug flag to try and find out.

Fallstar:
Ok, got it!
In https://github.com/JACoders/OpenJK/blob/e8b5c135eccb05ddae67e00ff944001f373fddd4/code/game/g_spawn.cpp#L1341 defaultStyles contains 14 correct entries, then only empty strings. Segfault occurs at line 1512 when it tries to get the length of an empty string.

I fixed it by setting all empty strings to `z`, like the normal style. The game now seems to work fine.
Can you think of a better way to fix it?

xycaleth:
That’s really weird! The length of an empty string is always (or should always be) 0. What map are you loading and do you have any mods?

Fallstar:
Yeah, my C++ is rusty so I don't really understand everything, it's not the value from the table that's directly used. It's set by` G_SpawnString` which uses pointer things I'm not used to.

I'm trying to load `yavin1` with no mods. Just trying to play the vanilla solo game. :)

vilhelmgray:
> That’s really weird! The length of an empty string is always (or should always be) 0. 

Those aren't empty strings: `{""}` is evaluating as `{0x0}`. The segmentation fault occurs since the call becomes `strlen(0x0)`.

mipek:
Thank you! I had the same issue and confirm that fixes it

vilhelmgray:
By the way, it looks like the original code works when compiled as **C** but not as **C++** so this is might be one of those small but significant differences between the two languages.

xycaleth:
[...]

PR #1025, opened 2019-11-11
Prevent segmentation fault when loading lightstyles

vilhelmgray:
The defaultStyle array only has 14 valid entries currently. When the
15th element is accessed, it returns a null-byte (not an empty string).
This commit prevents a segmentation fault by checking whether the
returned pointer from G_SpawnString is NULL before calling strlen.

Fixes: https://github.com/JACoders/OpenJK/issues/1021

### DIFF
--- a/code/game/g_spawn.cpp
+++ b/code/game/g_spawn.cpp
@@ -1337,96 +1337,6 @@ static	const char *defaultStyles[LS_NUM_STYLES][3] =
 		"zyxwvutsrqmlkjihgfedcba",
 		"aammbbzzccllcckkffyyggp"
 	},
-	{	// 14
-		"",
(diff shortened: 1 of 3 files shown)


vilhelmgray:
Does `defaultStyles` need to have `LS_NUM_STYLES` elements? If not, I can reduce the size of this array and hardcode the empty config strings; that should save space and speed up the process by eliminating the conditional checks for null pointers.

vilhelmgray:
This issue is likely the result of a regression in GCC that has fortunately been fixed for GCC version 9.3: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90947

Code changes to OpenJK may therefore not be necessary.

PR #1033, opened 2020-02-14
Remove FindSDL2 find-module, use sdl2-config.cmake instead

smcv:
*  Travis-CI: Move from Ubuntu 14.04 to 18.04

    Ubuntu 14.04 reached EOL for mainstream support in April 2019, and I'm
 about to add a dependency on a version of SDL newer than the one in
 14.04. Instead of going via 16.04, which only has about a year of
 support lifetime left itself, let's go directly to 18.04, currently the
 latest LTS release.

    Update build-dependencies accordingly.

* Remove FindSDL2 find-module, use sdl2-config.cmake instead

    This requires SDL >= 2.0.4.

    Since <https://bugzilla.libsdl.org/show_bug.cgi?id=2464> was fixed in
SDL 2.0.4, SDL behaves as a CMake "config-file package", even if it was
not itself built using CMake: it installs a sdl2-config.cmake file to
${libdir}/cmake/SDL2, which tells CMake where to find SDL's headers and
library, analogous to a pkg-config .pc file.

    As a result, we no longer need to copy/paste a "find-module package"
to be able to find a system copy of SDL >= 2.0.4 with find_package(SDL2).
Find-module packages are now discouraged by the CMake developers, in
favour of having upstream projects behave as config-file packages.

    This results in a small API change: FindSDL2 used to set SDL2_INCLUDE_DIR
and SDL2_LIBRARY, but the standard behaviour for config-file packages is
to set <name>_INCLUDE_DIRS and <name>_LIBRARIES. Use the CONFIG keyword
to make sure we search in config-file package mode, and will not find a
FindSDL2.cmake in some other directory that implements the old interface.

    In addition to deleting redundant code, this avoids some assumptions in
FindSDL2 about the layout of a SDL installation. The current libsdl2-dev
package in Debian breaks those assumptions; this is considered a bug
and will hopefully be fixed soon, but it illustrates how fragile these
assumptions can be. We can be more robust against different installation
layouts by relying on SDL's own CMake integration.

    When linking to a copy of CMake in a non-standard location, users can
now set the SDL2_DIR or CMAKE_PREFIX_PATH environment variable to point
to it; previously, these users would have used the SDL2DIR environment
variable. This continues to be unnecessary if using matching system-wide
installations of CMake and SDL2, for example both from Debian.

Mitigates: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=951087


ensiform:
Does this cause issue with local sdl2 on Windows (without having pkg config)?

smcv:
> Does this cause issue with local sdl2 on Windows (without having pkg config)?

If you're using the bundled libSDL (like the AppVeyor builds do), this commit doesn't change anything. This is the default in Windows builds, and I suspect it's what everyone actually does.

Some Windows users might be using a separate, system copy of libSDL (let's say you have OpenJK in `C:\games\OpenJK` and you want to use a SDL that you installed in `C:\libs\SDL2`). Previously, these users would have to compile with `SDL2DIR="C:\libs\SDL2"` or something. They should still be able to compile after this change, but using a slightly different environment variable, either `SDL2_DIR="C:\libs\SDL2"` or `CMAKE_PREFIX_PATH="C:\libs\SDL2"`. (I haven't tried this, but it's meant to work.)

code/CMakeLists.txt
```diff
@@ -89,9 +89,9 @@ if(BuildSPEngine OR BuildJK2SPEngine)
 			${OpenJKLibDir}/SDL2/include
 			)
 	else()
-		find_package(SDL2 REQUIRED)
-		set(SPEngineIncludeDirectories ${SPEngineIncludeDirectories} ${SDL2_INCLUDE_DIR})
-		set(SPEngineLibraries ${SPEngineLibraries} ${SDL2_LIBRARY})
+		find_package(SDL2 REQUIRED CONFIG)
```
smcv: > This requires SDL >= 2.0.4.

This breaks the Travis-CI build, because Travis-CI is using Ubuntu 14.04, which only has SDL 2.0.2.

I'll add a commit to bump it up to Ubuntu 18.04, which is the latest LTS, unless you particularly care about older versions? (16.04 just about has a new enough SDL.)

smcv: Branch updated. Let me know if you'd prefer 16.04 (but it only has about a year of support lifetime left so maybe not).

smcv:
I've fixed the Travis-CI failures. Please could someone take another look at this?

Issue #1038, opened 2020-04-30
out of bounds memory access in snd_mix.cpp S_PaintChannelFrom16


namtsui:
Reporting a bug? Please make sure you've given the following information - thanks!

**Operating system and version:**

OpenBSD amd64 
https://github.com/JACoders/OpenJK/commit/52030235f052772008d99e6ccb16de48e7ddb688

**Is this for single player or multiplayer?**

multiplayer

**Description of the bug (and if possible, steps to reproduce the bug):**

In multiplayer I created a game on Vjun Sentinel, got the Merr-Sonn PLX-2M behind the lifts near the starting spawn at lifts, and segfaulted either immediately or several seconds after firing it. This is somewhat hard to reproduce as most of the time I am able to fire without crashing.

```
(gdb) run
Starting program: /usr/local/share/JediAcademy/openjk 
[New thread 614481]
Thread 1 received signal SIGSEGV, Segmentation fault.
0x000006dce7350b97 in S_PaintChannelFrom16 (ch=0x6dce8177d00 <s_channels+920928>, sfx=0x6dce82f8890 <s_knownSfx+40832>, count=1024, sampleOffset=133859, bufferOffset=0) at /usr/ports/pobj/openjk-0.0.0.20200214/OpenJK-52030235f052772008d99e6ccb16de48e7ddb688/codemp/client/snd_mix.cpp:284
284			iData = sfx->pSoundData[ (int)ofst ];
(gdb) bt
#0  0x000006dce7350b97 in S_PaintChannelFrom16 (ch=0x6dce8177d00 <s_channels+920928>, sfx=0x6dce82f8890 <s_knownSfx+40832>, count=1024, sampleOffset=133859, bufferOffset=0) at /usr/ports/pobj/openjk-0.0.0.20200214/OpenJK-52030235f052772008d99e6ccb16de48e7ddb688/codemp/client/snd_mix.cpp:284
#1  0x000006dce7350ac9 in ChannelPaint (ch=0x6dce8177d00 <s_channels+920928>, sc=0x6dce82f8890 <s_knownSfx+40832>, count=1024, sampleOffset=133859, bufferOffset=0) at /usr/ports/pobj/openjk-0.0.0.20200214/OpenJK-52030235f052772008d99e6ccb16de48e7ddb688/codemp/client/snd_mix.cpp:353
[...]

PR #1040, opened 2020-05-21
Add bounds check to resolve buffer over-read

namtsui:
Fixes https://github.com/JACoders/OpenJK/issues/1038

ofst over-reads past the end of the pSoundData buffer. Add a bounds check
similar to S_CheckAmplitude() in codemp/client/snd_dma.cpp.

https://github.com/JACoders/OpenJK/commit/69800e894dff5ea9fcad22c26cc41c9bbf250216

### DIFF
--- a/codemp/client/snd_mix.cpp
+++ b/codemp/client/snd_mix.cpp
@@ -281,6 +281,9 @@ static void S_PaintChannelFrom16( channel_t *ch, const sfx_t *sfx, int count, in
 
 	for ( int i=0 ; i<count ; i++ )
 	{
+		// have we run off the end?
+		if ((int)ofst >= sfx->iSoundLengthInSamples)
+			break;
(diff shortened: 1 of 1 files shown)


xycaleth:
Thanks for the PR! I think a better place to fix this is here:
https://github.com/JACoders/OpenJK/blob/master/codemp/client/snd_mix.cpp#L445-L448

Before the code calls `S_PaintChannelFrom16` (from `ChannelPaint`)

Then we have the correct number of samples to add from the start and don't need to perform a comparison every sample loop

namtsui:
That seems like a better place to put it. Please test this to make sure I did not break the doppler effect when shooting a rocket launcher. It works in my testing with ten runs of shooting the rocket launcher.

xycaleth:
I'm unable to test right now but the code looks correct and I'll trust your testing went fine :)

PR #1042, opened 2020-06-03
GCC 10 Compliance

### LINKED ISSUE #1034: commit 5203023 breaks Arch Linux SDL2

Manjaro Linux 19.0.2

Build issue

commit 5203023 breaks Arch Linux build.  ${SDL2_INCLUDE_DIRS} and ${SDL2_LIBRARIES} return no value.

Expected SDL2 link libraries and includes to be available during compilation.  SDL.h not found.

As with any trouble-shooting, one can only say where the problem is when they know what solves it.  This patch solves the problem but I don't understand why.  I simply used the information from this unrelated github issue.
https://github.com/Palm-Studios/sh3redux/issues/134
```diff
diff --git a/code/CMakeLists.txt b/code/CMakeLists.txt
index 3c844d1e7..6f60c1d2e 100644
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -90,6 +90,8 @@ if(BuildSPEngine OR BuildJK2SPEngine)
 			)
 	else()
 		find_package(SDL2 REQUIRED CONFIG)
+		get_target_property(SDL2_INCLUDE_DIRS SDL2::SDL2 INTERFACE_INCLUDE_DIRECTORIES)
+		get_target_property(SDL2_LIBRARIES SDL2::SDL2 IMPORTED_LOCATION_NOCONFIG)
 		set(SPEngineIncludeDirectories ${SPEngineIncludeDirectories} ${SDL2_INCLUDE_DIRS})
 		set(SPEngineLibraries ${SPEngineLibraries} ${SDL2_LIBRARIES})
 	endif()
diff --git a/codemp/CMakeLists.txt b/codemp/CMakeLists.txt
index a55ddf789..be3765cd2 100644
--- a/codemp/CMakeLists.txt
+++ b/codemp/CMakeLists.txt
@@ -397,6 +397,8 @@ if(BuildMPEngine)
 			)
 	else()
 		find_package(SDL2 REQUIRED CONFIG)
+		get_target_property(SDL2_INCLUDE_DIRS SDL2::SDL2 INTERFACE_INCLUDE_DIRECTORIES)
+		get_target_property(SDL2_LIBRARIES SDL2::SDL2 IMPORTED_LOCATION_NOCONFIG)
 		set(MPEngineIncludeDirectories ${MPEngineIncludeDirectories} ${SDL2_INCLUDE_DIRS})
 		set(MPEngineLibraries ${MPEngineLibraries} ${SDL2_LIBRARIES})
 	endif()
```

cagelight:
1. GCC 10 forbids multiple declarations of global variables, Com_Printf and Com_Error throw compiler errors as a result.
2. Fixes #1034

### DIFF
--- a/codemp/cgame/cg_main.c
+++ b/codemp/cgame/cg_main.c
@@ -26,6 +26,10 @@ along with this program; if not, see <http://www.gnu.org/licenses/>.
 #include "cg_local.h"
 
 #include "ui/ui_shared.h"
+
+NORETURN_PTR void (*Com_Error)( int level, const char *error, ... );
+void (*Com_Printf)( const char *msg, ... );
+
 // display context for new ui stuff
 displayContextDef_t cgDC;
 
--- a/codemp/game/g_main.c
+++ b/codemp/game/g_main.c
@@ -29,6 +29,9 @@ along with this program; if not, see <http://www.gnu.org/licenses/>.
 #include "bg_saga.h"
 #include "b_local.h"
 
+NORETURN_PTR void (*Com_Error)( int level, const char *error, ... );
+void (*Com_Printf)( const char *msg, ... );
+
 level_locals_t	level;
 
 int		eventClearTime = 0;
--- a/codemp/qcommon/q_shared.h
+++ b/codemp/qcommon/q_shared.h
@@ -687,8 +687,8 @@ qboolean Info_NextPair( const char **s, char *key, char *value );
 
 // this is only here so the functions in q_shared.c and bg_*.c can link
 #if defined( _GAME ) || defined( _CGAME ) || defined( UI_BUILD )
-	NORETURN_PTR void (*Com_Error)( int level, const char *error, ... );
-	void (*Com_Printf)( const char *msg, ... );
+	extern NORETURN_PTR void (*Com_Error)( int level, const char *error, ... );
+	extern void (*Com_Printf)( const char *msg, ... );
 #else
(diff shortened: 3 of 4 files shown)


cagelight:
Seems like CMake 3.0. I've been doing it this way for years and it seems to be the recommended way to use libraries now.

code/CMakeLists.txt
```diff
@@ -90,8 +90,8 @@ if(BuildSPEngine OR BuildJK2SPEngine)
 			)
 	else()
 		find_package(SDL2 REQUIRED CONFIG)
-		set(SPEngineIncludeDirectories ${SPEngineIncludeDirectories} ${SDL2_INCLUDE_DIRS})
-		set(SPEngineLibraries ${SPEngineLibraries} ${SDL2_LIBRARIES})
+		set(SPEngineIncludeDirectories ${SPEngineIncludeDirectories})
+		set(SPEngineLibraries ${SPEngineLibraries} SDL2::SDL2)
```
xycaleth, line 94: Thanks for the PR!

While the syntax is available since CMake 3.0, SDL2 was only updated this March to have the `SDL2::SDL2` target (http://hg.libsdl.org/SDL/diff/ef2d5382a887/sdl2-config.cmake.in). We want to be supporting older versions of SDL2 so we can't accept this change right now unfortunately!

xycaleth:
appveyor is kind of broken right now so no need to worry about those.

PR #1055, opened 2020-11-07
Fixed issue with toolchain not being detected in Docker build

everon:
Fixes an issue with CMake not resolving the correct path, it isn't clear why this isn't working.
Replacing it with an absolute path (via realpath) allows this to compile.


xycaleth (approved):
Looks good! Thanks for the contribution

PR #1056, opened 2020-11-11
Fix singleplayer crashes

z33ky:
I encountered and debugged these two crashes playing the Jedi Outcast singleplayer on Linux.
I've also applied the codeJK2 fix to code(JK3) since I would bet it can happen there as well.

### DIFF
--- a/code/cgame/cg_credits.cpp
+++ b/code/cgame/cg_credits.cpp
@@ -214,13 +214,11 @@ static const char *GetSubString(std::string &strResult)
 // ...
 static bool SortBySurname(const StringAndSize_t &str1, const StringAndSize_t &str2)
 {
-	std::string::const_reverse_iterator rbegin1 = str1.str.rbegin();
-	std::string::const_reverse_iterator rbegin2 = str2.str.rbegin();
+	std::string::const_reverse_iterator rstart1 = std::find_if(str1.str.rbegin(), str1.str.rend(), isspace);
+	std::string::const_reverse_iterator rstart2 = std::find_if(str2.str.rbegin(), str2.str.rend(), isspace);
+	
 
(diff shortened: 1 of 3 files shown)


code/cgame/cg_credits.cpp
```diff
@@ -442,7 +437,7 @@ void CG_Credits_Init( const char *psStripReference, vec4_t *pv4Color)
 					{
 						// sort entries RHS dotted entries by alpha...
 						//
-						qsort(&CreditLine.vstrText[0], CreditLine.vstrText.size(), sizeof(CreditLine.vstrText[0]), SortBySurname);
+						std::sort( CreditLine.vstrText.begin(), CreditLine.vstrText.end(), SortBySurname );
```
z33ky, line 440: Compared to the previous code, this will additionally require the header <algorithm>.
It seems headers are mostly not included in this implementation file (i.e. <string> or <vector> isn't included directly). It compiled on my machine, so I left the includes as-is. Let me know if I need to add that somewhere to accommodate this change.

code/rd-vanilla/tr_surface.cpp
```diff
@@ -1617,6 +1617,10 @@ static void RB_SurfaceLathe()
 	{
 		lod = 4;
 	}
+	if ( lod < 1 )
+	{
+		lod = 1;
+	}
```
xycaleth, line 1623: What does this change? Seems unrelated to the other changes :)

xycaleth: Actually never mind, I see the problem

xycaleth (approved):

xycaleth:
Thanks for the contribution @z33ky!

PR #1057, opened 2020-11-12
Added the libcurl on the Docker runner image

alexaugustobr:
Popular mods may require this lib to run properly, like movie battles 2 mod. 
With this change this docker image can used seamlessly, in order to create mb2 servers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ FROM ubuntu:18.04
 # Install utilities and libraries
 RUN dpkg --add-architecture i386 &&\
 	apt-get -q update &&\
-	DEBIAN_FRONTEND="noninteractive" apt-get -q install -y -o Dpkg::Options::="--force-confnew" --no-install-recommends socat libstdc++6 libstdc++6:i386 zlib1g zlib1g:i386 &&\
+	DEBIAN_FRONTEND="noninteractive" apt-get -q install -y -o Dpkg::Options::="--force-confnew" --no-install-recommends socat libstdc++6 libstdc++6:i386 zlib1g zlib1g:i386 libcurl4 libcurl4:i386 &&\
 	rm -rf /var/lib/apt/lists/*
 
 # Copy binaries and scripts


alexaugustobr:
Hey @xycaleth can you reviewing help with that?

xycaleth:
Maybe I'm missing something, but I don't see how this will allow other mods to work seamlessly In the Docker image? To run MB2 in this image, the image would need to be changed to compile the MB2 source code in addition or instead of the OpenJK code. If that needs to happen then the packages to be installed would need to be changed anyway.

alexaugustobr:
No, with this change MB2 source code doesn't need to be compiled or something like that. The problem is that some MB2 library (.so file) needs libcurl4, and looks like it's not come on the OS any longer by default.
MB2 source code it's not publicly available.

xycaleth:
Ah yes, I forgot MB2 hasn't released the source. How do you plan to add `+set fs_game MBII` to the server command line and to make the MBII assets and `.so` file available to the server?

I really like the idea of being able to re-use the same image for any mods, but I'm cautious of pre-installing all of the package dependencies that any existing (or future) mod will require.

alexaugustobr:
Well it works simple, you need to tell the game where is the home and base path.

I'm using docker compose, my yml looks like that:
```
services: 
  mbii-server:
    image: jacoders/openjk
    restart: always
    ports:
      - "29070:29070/udp"
    volumes:
      - game-files:/opt/openjk/basepath
      - server-log-config:/opt/openjk/homepath
    environment: 
      - OJK_MOD=MBII
      - OJK_OPTS=+set dedicated 2 +set net_port 29070 +set fs_cdpath /opt/openjk/cdpath +set fs_basepath /opt/openjk/basepath +set fs_homepath /opt/openjk/homepath +set fs_game MBII

```
Using these volumes, the just place the "MBII" folder alongside the "base" folder on "game-files" volume, and the mod will run.
The "server-log-config" volume is for server.cfg and log files. 

About the dependencies, you are right, but I think we can open an exception for the MB2.

z33ky:
Why not write a simple Dockerfile that contains the additional dependencies and use that?
```
FROM jacoders/openjk
RUN apt-get update && \
    apt-get -y install libcurl4 libcurl4:i386 && \
    rm -rf /var/lib/apt/lists/*
```
You still automatically get any updates made to the openjk Docker image.

alexaugustobr:
Yes, @z33ky this can be done, I'm already running a custom version of this Dockerfile. So you guys can close this pull if does not look useful. Thanks.

xycaleth:
> Why not write a simple Dockerfile that contains the additional dependencies and use that?
> 
> ```
> FROM jacoders/openjk
> RUN apt-get update && \
>     apt-get -y install libcurl4 libcurl4:i386 && \
>     rm -rf /var/lib/apt/lists/*
> ```
> 
> You still automatically get any updates made to the openjk Docker image.

I much prefer this solution than changing the existing Dockerfile. Thanks @z33ky 👍 

I'll go ahead and close this PR.

PR #1064, opened 2020-12-09
Rend2 shader constants new

Almightygir:
Added a cloth BRDF, fixed up some things in diffuse, added some dummy #ifs for us to define things properly later.


xycaleth:
Wrong fork 😄

PR #1082, opened 2021-03-22
Fixed assertions in non-debug environment

KiralyCraft:
Some assertions were done on variables defined in a debug mode, which were inaccessible in non-debug mode and caused the compilation to fail. I added checks to only include the assertions in debug mode.


xycaleth (approved):

xycaleth:
Thanks for the contribution!

PR #1089, opened 2021-05-01
 Added CMakeSettings.json file.

j-horner:
Allows easy build with Visual Studio - CMake integration.
Debug and Release configurations added.
installRoot set to `build/install/`


ensiform:
Curious why only x64 builds when the preferred for windows builds is 32?

j-horner:
Partly personal habit of just building x64 and partly because I didn't realise 32 was the preferred.

I'll add x86 and some RelWithDebInfo configurations too.

ensiform:
> Partly personal habit of just building x64 and partly because I didn't realise 32 was the preferred.
> 
> I'll add x86 and some RelWithDebInfo configurations too.

Just because there are no mods for 64-bit out there, and many people still play old mods like JA+, 64-bit is unavailable for them.
Also, it adds no benefit to JKA engine to do so.

xycaleth (commented):
I'm on the fence about approving this since we've so far managed to not have any text editor or IDE specific files in the repo :/

j-horner:
> we've so far managed to not have any text editor or IDE specific files in the repo
That's understandable. I was going to close the PR but I saw that CMake added a new feature called Presets, which is basically the sort of JSON file except that it can be parsed by CMake directly and isn't IDE specific.

https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#build-preset

I added some x86 configurations and it seems to work well with VS code. VS only supports it under preview at the moment so I haven't test that. I'm not sure about other IDEs.

`CMakeUserPresets.json` is apparently intended for local developer settings so I added it to .gitignore and made a simple template for overriding `CMAKE_INSTALL_PREFIX`

PR #1090, opened 2021-05-04
build: pass -fsigned-char on clang

tobhe:
Powerpc and ARM default to unsigned char which causes an immediate crash when starting the game.
On gcc this is prevented by passing -fsigned-char in CFLAGS and CXXFLAGS. This change does the same for clang.


bluerise (approved):

xycaleth:
Out of curiosity, where is it crashing specifically when you try to start the game?

tobhe:
The actual segfault happens in `CFxScheduler::AddScheduledEffects()`: https://github.com/JACoders/OpenJK/blob/d2ed03a016f54917bc1dcf06bcc59d3f674fb533/code/cgame/FxScheduler.cpp#L1403

The reason seems to be that the previous check in 
https://github.com/JACoders/OpenJK/blob/d2ed03a016f54917bc1dcf06bcc59d3f674fb533/code/cgame/FxScheduler.cpp#L1381
doesn't work because `mBoltNum` as a char is unsigned and the compiler decides that this can never be true.

xycaleth (approved):
This looks good to me. Maybe in the future we can stop relying on char signedness but this is a good quick fix.

PR #1093, opened 2021-05-14
Disabled warning in q_shared.h

j-horner:


xycaleth (approved):
Thanks for the contribution!

PR #1094, opened 2021-05-18
Fixed some integer conversion warnings.

j-horner:
Warnings present when compiling with clang-cl 11.0


j-horner:
Probably too many changes for too little benefit to consider merging this but I started so I thought I'd persevere...

There shouldn't be any functional changes, just explicit casts highlighting some implicit conversions. Mostly implicit integer conversions.

PR #1095, opened 2021-06-09
Fixed possible undefined behaviour

j-horner:
Similiar to that in https://github.com/JACoders/OpenJK/commit/92059a6bdd9a7a256e0a07d550c5fc4db4ddb8f4


Razish:
Can this be rebased on top of master for the automated builds to run? I'm suspect of this working on MSVC.
If it works fine, it's a :+1: from me

bartrpe (approved):
@Razish it actually builds on MSVC, I just tested on whatever is the newest MSVC version built into Visual Studio Enterprise 2022

Daggolin (commented):
The check seems to be correct, but does this actually occur? Multiplayer doesn't allow for non-glm NPCs. And if I remember correctly all glm NPCs should pass through `SetupGameGhoul2Model`, which should `ERR_DROP` on invalid `localAnimIndex`.

PR #1103, opened 2021-09-10
Upgraded Windows Libraries - SDL2 > 2.0.13 FullscreenFix

Nightshades1:
libjpeg 9a -> 9d
zlib 1.2.8 -> 1.2.11
libpng 1.6.16 -> 1.6.37
SDL2 2.0.12 -> 2.0.16

Fixed fullscreen issue in recent SDL2 version 2.0.14 and above due to SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS being set to 0 by default.


ArminiusTux:
Hi, is there a reason for not going the FULL mile?

```
libjpeg 9a -> 9e
zlib 1.2.8 -> 1.3
libpng 1.6.16 -> 1.6.40
OpenAL 0.9.9.5 -> 1.23.1
SDL2 2.0.12 -> 2.28.5
```
grafik

taysta:
> Hi, is there a reason for not going the FULL mile?
> 
> ```
> libjpeg 9a -> 9e
> zlib 1.2.8 -> 1.3
> libpng 1.6.16 -> 1.6.40
> OpenAL 0.9.9.5 -> 1.23.1
> SDL2 2.0.12 -> 2.28.5
> ```
> 
> grafik

Did you have it working across Linux, Mac and WIndows? You should make a pull request for these changes.

ensiform:
Upgrading OpenAL is ill advised because the included proprietary dll in the vanilla game isn't forward+backwards compatible. Iirc if you replace the newer version dll in the folder jamp/jasp will crash so you can't drop openjk in the same folder as the original and play openjk and original in the same install.

ArminiusTux:
> Did you have it working across Linux, Mac and WIndows? You should make a pull request for these changes.

Point taken 😉



> Upgrading OpenAL is ill advised because the included proprietary dll in the vanilla game isn't forward+backwards compatible. Iirc if you replace the newer version dll in the folder jamp/jasp will crash so you can't drop openjk in the same folder as the original and play openjk and original in the same install.

Interesting, so #1131 has to deal with the OpenAL situation as well then.

ensiform:
Or don't support openal on arm at all. It's unsupported everywhere else, including x64 windows.

Nightshades1:
Hi guys, i kept having notificatin on my mail 😁
I made that at the time i was playing Movie Battles II but it was generic to OpenJK anyway, i thought it was a good idea to upgrade the libraries.

Do note that when i sent that pull request, it was mostly only for Windows (the libraries upgrade), I can imagine that if you are on linux or even on windows, you would use cmake for it (like the code we have to paste if we use vcpkg on Windows)

But if you plan to use Higher SDL Version, then don't forget about the 
```"SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS"``` flag or you'll have issue with the fullscreen.

PR #1109, opened 2021-11-02
Fix crash or freeze when there is too much text visible in the console

aufau:
Game was freezing or crashing when there were too many characters visible at the same time in the console. Easy to reproduce with high resolution display.
Commit messages describe what was causing crashes and what had to be done.

After changes excessive characters are not printed but at least client remains functional, console prompt and latest console lines are visible:

obraz

Tested by @withRhodes


bartrpe (approved):
Checked that out on 3840x1080 (two screens merged) and it seems alright, as in - no crashes for me.

ensiform:
Replaced in another commit

PR #1116, opened 2022-03-08
Fixed some warnings

The-New-Order:


code/game/g_active.cpp
```diff
@@ -2904,8 +2904,8 @@ qboolean G_CheckClampUcmd( gentity_t *ent, usercmd_t *ucmd )
 							damage = Q_irand( 3, 5 );
 						}
 					}
-					if ( ent->client->ps.legsAnimTimer <= 1150
-						&& ent->client->ps.legsAnimTimer > 10500 )
+					if ( ent->client->ps.legsAnimTimer > 1150
```
ikopysov, line 2907: It seems this conflicts with #1119. I'm not sure which one is correct

tobhe: If you look at the other checks in the same function you will see that most are checking for a 100ms time window such as:
```
if ( ent->client->ps.legsAnimTimer <= 3150
    && ent->client->ps.legsAnimTimer > 3050 )
```
or
```
if ( ent->client->ps.legsAnimTimer <= 5700
    && ent->client->ps.legsAnimTimer > 5600 )
```
which is why I assumed mine would be the logical fix here. I could also be wrong though.

bartrpe (approved):
This fixes obvious typos in the code that messes up stuff. I think it should be merged, it's relatively straightforward

PR #1119, opened 2022-05-06
Fix typo that causes comparison to always evaluate to false.

tobhe:
Looking at the other checks in this code I think it is safe to assume that this was meant to be 1050.


ensiform (approved):
Looks fine to me. Everything else is 4 digit timer checks but this was incorrectly 5 digits.

Razish (approved):

PR #1125, opened 2022-09-02
[MP] seed RNG in modules

Razish:
This should fix unwanted side-effects introduced by 2bcf85c.
Each module has its own copy of the initial RNG seed (`holdrand`), but only the engine's copy was being updated on init.
There were many effects of this, notably the "random" spawns on a `map_restart` were actually consistent, since the game module gets reloaded.


bartrpe (approved):
works for me

bartrpe:
how's the status on this one? i think no new developments have been made and this version can be pushed?

SomaZ (approved):
Considering the lengthy discussion we had on discord reagarding this, I think this is fine for merging. This is the closest we can get with random numbers compared to the original executable considering different build systems. Randoms are called all over the place, also with user interactions, so theres no pattern that can be recreated for the random values.

Daggolin (approved):
As nobody found any indications of easily recognizable patterns and as the vanilla modules seem to have not used the `bg_lib.c` version of `rand` there is probably no need to revert the old commit and this should be enough to counter the mentioned side effects.

To reference the mentioned discussion and given the `every compile feels different` discussions over the years I can understand why someone would want to revert the mentioned commit, but I suppose if someone is afraid of hidden side effects it's easier to recompile modules from a codebase closer to the vanilla code than to try and undo the cleanup made in OpenJK.

So let's use this PR. If something comes up in the future the two commits could still be reverted.

bartrpe:
> Considering the lengthy discussion we had on discord reagarding this, I think this is fine for merging. This is the closest we can get with random numbers compared to the original executable considering different build systems. Randoms are called all over the place, also with user interactions, so theres no pattern that can be recreated for the random values.

Honestly you made a point that I think is often overlooked here, sometimes it feels like people feel pushing some new thing to repo is a lifetime commitment :P

PR #1131, opened 2022-12-19
Enable building for Windows ARM64

JunielKatarn:
Allows building natively for Windows/ARM64 devices (Surface Pro X, Volterra, etc).


ArminiusTux:
Greetings, just wanted to drop my experience in compiling the MP portion on WinARM64.
The above "__rdtsc workaround" really gets the other two executables, `openjk.x86_64.exe` & `openjkded.x86_64.exe` compiled, but they crash outright.

The [SP.exe](https://github.com/ArminiusTux/WinARM64-binaries/releases/tag/OpenJK) is not exactly rock solid, but absolutely playable.

ArminiusTux:
Is there any way to kick this can / pull request further down the road?

code/CMakeLists.txt
```diff
@@ -71,18 +71,24 @@ if(BuildSPEngine OR BuildJK2SPEngine)
...
 			set(SPEngineLibraries
 				${SPEngineLibraries}
 				${OpenJKLibDir}/SDL2/lib/x64/SDL2main.lib
 				${OpenJKLibDir}/SDL2/lib/x64/SDL2.lib
 				)
+		elseif(Architecture STREQUAL "arm64")
+			set(SPEngineLibraries
+				${SPEngineLibraries}
+				${VcpkgDir}/installed/arm64-windows/lib/manual-link/SDL2main.lib
+				${VcpkgDir}/installed/arm64-windows/lib/SDL2.lib
```
Razish, line 90: I don't love having vendored versions for everything _except_ arm64-windows.
I think my preference would be vendored versions for all, with optional sourcing from (system) package manager. I also don't love how much bloat and churn that entails. Maybe arm64-windows is a special-enough case to push for vcpkg.

This isn't a blocker, just provoking discussion so we can form a stance here intentionally.

PR #1134, opened 2023-01-16
Add Support For Modder GLA - SP

DT85:
So modders can create their own GLA with their own anims for NPCs, not sure about if the player has access as well. Is indexed after the map GLA if it's present, otherwise indexed after the base GLA. Added 50 modder anims to the anim headers, "BOTH_MOD_XX".


mrwonko:
As far as I can tell, there are two new changes being introduced here:

1. Loading additional animations from `models/players/_humanoid/_humanoid_mod.gla`
2. The addition of BOTH_MOD_N animations

Loading animations from a single additional file does not provide modders with many new options. You still can't combine different animation mods, so you might as well just overwrite the original _humanoid.gla, making this change unnecessary. I generally like the idea of letting modders selectively add animations, but I think it should be possible to combine such mods.

As for the addition of BOTH_MOD_XY animations, I don't really see the benefit compared to the existing BOTH_CIN_XY? Those are already available for level-specific animations. I suppose the difference is that these are not meant to be level specific? But with a limited set of pre-defined generic names, there's bound to be clashes between mods using these. It would be more useful if mods could add new animations with arbitrary names, though I'm not sure if that's in the scope of OpenJK.

Could you outline some example use cases for how to take advantage of these new features? What would we use them for?

PR #1138, opened 2023-02-01
game: fix overflow on too many spawnpoints

### LINKED ISSUE #1137: SP crashes on certain MP maps
Reportedly, the MP map <https://mrwonko.de/jk3files/Jedi%20Academy/Maps/Free%20For%20All/82455/> crashes when loaded in SP with no clear error message.

It's fine for SP to not support MP maps, but there should be an error message explaining the problem, not a crash.

mrwonko:
Previously, the game would typically just crash due to stack corruption after a buffer overflow. Now, we detect the overflow, and show an error instead.

Can be tested using <https://mrwonko.de/jk3files/Jedi%20Academy/Maps/Free%20For%20All/82455/>.

Fixes #1137

### DIFF
--- a/code/game/g_client.cpp
+++ b/code/game/g_client.cpp
@@ -239,6 +239,9 @@ gentity_t *SelectRandomDeathmatchSpawnPoint( team_t team ) {
 	spot = NULL;
 
 	while ((spot = G_Find (spot, FOFS(classname), "info_player_deathmatch")) != NULL) {
+		if (count >= MAX_SPAWN_POINTS) {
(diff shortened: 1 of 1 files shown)


SomaZ:
I'd prefer it to break the while loop and show a warning in the console instead of erorring out completely.

bartrpe:
> I'd prefer it to break the while loop and show a warning in the console instead of erorring out completely.

yeah, I think that's the best direction to take

ensiform:
Wouldn't it be better to copy the MP code?

PR #1153, opened 2023-06-25
Fix hanging cutscenes on fast machines

### LINKED ISSUE #1150: Cutscenes failing to play correctly
I'm experiencing an issue in Jedi Academy on the cutscene where Luke is addressing the students after the first level. The game does not crash but nothing happens. No animation, no camera movement, no speech. You have to skip the cutscene to progress. This happens in both vanilla Jedi Academy and with the latest OpenJK.

I've found a number of reports of the same issue online:
https://www.gog.com/forum/star_wars_jedi_knight_jedi_academy/first_jedi_temple_cutscene_stuck
https://www.gog.com/forum/star_wars_jedi_knight_jedi_academy/game_freezing
https://steamcommunity.com/app/6020/discussions/0/3276942370878082914/
https://steamcommunity.com/app/6020/discussions/0/3201493200082464233/
https://steamcommunity.com/app/6020/discussions/0/1696048426810765937/
https://www.reddit.com/r/jediknight/comments/ek9mkq/jedi_academy_yavin_iv_cutscene_glitch/

Debugging this issue was a pain because debug builds of OpenJK played the cutscene just fine, but I got there in the end. It turns out that the issue occurs when the level loads too fast.

The game server spawns entities behind the scenes while the loading bar is moving. When the loading bar reaches the end, the client connects to the server and the player is placed in the world. If this occurs before all the entities are loaded then any script that references any of the missing entities will fail to load fully, leaving the chain of sequencer tasks broken.

When the server starts (at the beginning of the loading screen), the server time is set to 1000 milliseconds:
https://github.com/JACoders/OpenJK/blob/90e8005b0cef3ce3405ad3f0011a204ca78e4c37/code/server/sv_init.cpp#L288

It then runs frames at 1000, 1100, 1200 and 1300 milliseconds to "warm up" the sever:
https://github.com/JACoders/OpenJK/blob/90e8005b0cef3ce3405ad3f0011a204ca78e4c37/code/server/sv_init.cpp#L313

Meanwhile, the entity spawner's `nextthink` time is set to 1350 and its `e_ThinkFunc` is set to `thinkF_NPC_Spawn_Go`:
https://github.com/JACoders/OpenJK/blob/90e8005b0cef3ce3405ad3f0011a204ca78e4c37/code/game/NPC_spawn.cpp#L2048-L2049

The first non-warmup frame occurs at 1400+(1/sv_fps) milliseconds, which with the default settings is at 1450. This triggers the think function, which finally bumps the `nextthink` time along by 100 milliseconds to 1550 and sets `e_ThinkFunc` to `thinkF_NPC_Begin`:
https://github.com/JACoders/OpenJK/blob/90e8005b0cef3ce3405ad3f0011a204ca78e4c37/code/game/NPC_spawn.cpp#L1758-L1759

If this next time is reached before the client connects to the server (the loading bar reaches the end) then everything runs smoothly. The `NPC_Begin` function has been run on each entity and so they can all be looked up correctly by name, specifically here:
https://github.com/JACoders/OpenJK/blob/90e8005b0cef3ce3405ad3f0011a204ca78e4c37/code/icarus/Sequencer.cpp#L692-L693

On my laptop, and in a release build, the client connects to the server at 1400 milliseconds. With RelWithDebugInfo I get around 1550 to 1600 on the first load of the cutscene, meaning it works, and then 1450 to 1500 on reloads (probably because some assets are still in memory?), meaning it fails.

I'm unsure what the best way to fix this bug is. The easiest way would probably be to increase the number of warm up frames or hardcode a delay to the client connection until after 1550 milliseconds. Alternatively, the Icarus code could be changed to postpone the lookup of entities until the task is processed, but  I wouldn't know where to start with that.

lawadr:
On machines that can load levels faster than the hardcoded entity spawning delay, the Icarus sequencer could reference non-existing entities and therefore fail to chain events to progress cutscenes.

This change increases the number of warmup frames by 1 to push the possible client-server connection time to after all relevant entities have been spawned.

Fixes #1150

### DIFF
--- a/code/server/sv_init.cpp
+++ b/code/server/sv_init.cpp
@@ -311,7 +311,7 @@ void SV_SpawnServer( const char *server, ForceReload_e eForceReload, qboolean bA
 	SV_InitGameProgs();
 
 	// run a few frames to allow everything to settle
-	for ( i = 0 ;i < 3 ; i++ ) {
+	for ( i = 0 ;i < 4 ; i++ ) {
 		ge->RunFrame( sv.time );
 		sv.time += 100;
 		re.G2API_SetTime(sv.time,G2T_SV_TIME);


ensiform:
After all that checking and investigating only adding 1 additional frame was sufficient? Crazy.

lawadr:
> After all that checking and investigating only adding 1 additional frame was sufficient? Crazy.
> 
> I do know that Enemy Territory uses 6 frames but I never really knew why. Maybe their mod code scripting depends on that too. But it's MP only. Perhaps Rtcw SP also does.

Very.

There's a few constants defined in the source, such as `START_TIME_REMOVE_ENTS`, whose purpose seems to be to stagger things enough to just make everything work. Initialisation events seem to happen in a series of "waves" separated by 50 to 100ms. If the client connects too early then you'll probably hit an issue of some kind.  I guess you just need to make sure the warmup frames are sufficient for the number of these "waves" in a game.

Obviously computers were nowhere near fast enough back then to ever hit this issue. I think what finally did it was SSD speeds. I was loading a level in 0.4 seconds.

Razish (dismissed):

Daggolin (approved):
If I recall correctly the discussion mentioned by @Razish concluded that this pull request should work for vanilla maps, but there were concerns about possible custom missions or other custom mods.

As any other approach to this issue is likely to take some time and changes to a lot more places of code I would suggest to merge this for now, because several users have reported that they couldn't finish the game, because they would get stuck due to this issue over the past few months.

If someone is concerned about possible side effects one could add a cvar to specify the amount of warmup frames and default it to `4` so the change can easily be reverted by users for testing purposes or extended if a custom mission requires even more warmup frames.

Razish (approved):

PR #1154, opened 2023-07-26
Update gsl-lite

cagelight:
The ancient version of gsl-lite currently on the repo does not compile on modern setups, so I have updated the gsl-lite file and taken care of all of the things that have been renamed since the older file. Seemed like just a mass find/replace operation more or less.

### DIFF
--- a/code/cgame/FxScheduler.cpp
+++ b/code/cgame/FxScheduler.cpp
@@ -524,7 +524,7 @@ int CFxScheduler::ParseEffect( const char *file, const CGPGroup& base )
 
 	for( const auto& primitiveGroup : base.GetSubGroups() )
 	{
-		static std::map< gsl::cstring_view, EPrimType, Q::CStringViewILess > primitiveTypes{
+		static std::map< gsl::cstring_span, EPrimType, Q::CStringViewILess > primitiveTypes{
 			{ CSTRING_VIEW( "particle" ), Particle },
 			{ CSTRING_VIEW( "line" ), Line },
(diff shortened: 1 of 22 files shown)


MichaelColonel:
I've tested this PR on Linux, both Outcast and Academy single player, Release build with gcc-13.2. No problems with compilation and execution.

Before this PR most of gsl-lite errors on Linux ,in my case, were because of undefined `std` namespace within `gsl` namespace.

Simple fix for me was to replace `std::` to `::std::`.
For example:
`std::reverse_iterator< iterator >  reverse_iterator;` -> `::std::reverse_iterator< iterator > reverse_iterator;`

Razish (approved):

bartrpe (approved):
Gave it a test on Debian 12.1 with the newest GCC version and didn't encounter any errors building.

Daggolin (approved):
> Gave it a test on Debian 12.1 with the newest GCC version and didn't encounter any errors building.

As far as I know Debian 12 is still on gcc 12. Using gcc12 works for me on the current `master` branch. Using gcc13 or clang16 did not. With this pull request they work as well.

The sha256sum of `lib/gsl-lite/include/gsl/gsl-lite.h` matches the current https://raw.githubusercontent.com/gsl-lite/gsl-lite/master/include/gsl/gsl-lite.hpp, so it seems to still be the latest version.

Seems like most of the changes replace `cstring_view` with `cstring_span`, because `cstring_view` seems to have been renamed to `cstring_span` back in 2015.

PR #1155, opened 2023-08-08
Fix gsl-lite compilation errors

MichaelColonel:
Similar to #1154 but fix only gsl-lite header. Compilation process has been tested on Linux with gcc-13.2 compiler.


smcv:
Resolves: #1159

Issue #1159, opened 2023-09-12
Included gsl-lite fails to compile with g++-13 13.2.0


smcv:
**Operating system and version:**
Debian testing/unstable, with g++ 13.2.0 as default C++ compiler.

**Is this for single player or multiplayer?**
yes :-)

**Description of the bug (and if possible, steps to reproduce the bug):**
To reproduce: build from source, with JK2 enabled (and I'm using `CMAKE_BUILD_TYPE=Debug UseInternalZlib=0 BuildTests=ON`, if that matters).

I get lots of C++ errors from `gsl-lite.h`, here are the first few:

```
In file included from /home/smcv/src/openjk/shared/qcommon/safe/gsl.h:4,
                 from /home/smcv/src/openjk/shared/qcommon/safe/files.h:7,
                 from /home/smcv/src/openjk/shared/qcommon/safe/files.cpp:1:
/home/smcv/src/openjk/lib/gsl-lite/include/gsl/gsl-lite.h:479:18: error: ‘reverse_iterator’ in namespace ‘gsl::std’ does not name a template type
  479 |     typedef std::reverse_iterator< iterator >                reverse_iterator;
      |                  ^~~~~~~~~~~~~~~~
/home/smcv/src/openjk/lib/gsl-lite/include/gsl/gsl-lite.h:480:18: error: ‘reverse_iterator’ in namespace ‘gsl::std’ does not name a template type
  480 |     typedef std::reverse_iterator< const_iterator >          const_reverse_iterator;
      |                  ^~~~~~~~~~~~~~~~
/home/smcv/src/openjk/lib/gsl-lite/include/gsl/gsl-lite.h:483:27: error: ‘iterator_traits’ in namespace ‘gsl::std’ does not name a template type
  483 |     typedef typename std::iterator_traits< iterator >::difference_type difference_type;
      |                           ^~~~~~~~~~~~~~~
/home/smcv/src/openjk/lib/gsl-lite/include/gsl/gsl-lite.h:483:42: error: expected unqualified-id before ‘<’ token
  483 |     typedef typename std::iterator_traits< iterator >::difference_type difference_type;
      |                                          ^
/home/smcv/src/openjk/lib/gsl-lite/include/gsl/gsl-lite.h:493:47: error: expected ‘)’ before ‘,’ token
  493 |     gsl_constexpr14 array_view( std::nullptr_t, size_type size )
      |                               ~               ^
      |                                               )
[...]

PR #1160, opened 2023-09-12
gsl: Fully qualify std:: types as being in the global namespace

smcv:
With g++ 13, if we don't do this for names appearing inside `namespace gsl {}`, then they're interpreted as being `gsl::std::unique_ptr` and so on.

Resolves: https://github.com/JACoders/OpenJK/issues/1159


smcv:
Never mind, duplicate of #1155.

PR #1165, opened 2023-09-16
Multiplayer RoffSystem Fixes

Daggolin:
This pull request fixes two multiplayer bugs related to the RoffSystem:
 1. The RoffSystem used the wrong time value (engine time rather than module time) since the times were separated in 92e12a9e9b19525c1f589f1f6779e9397b6bfd26. This caused playback to be incorrect after the first mapchange. The same bug was also introduced by jk2mv and has been fixed the same way in https://github.com/mvdevs/jk2mv/commit/11a108ed4c495b0808f9f15f72f1668149cc4b49.
 2. The RoffSystem used `long` instead if `int` for the `mVersion` header field, causing 64-bit versions to be unable to parse roff files and playing them back.

### DIFF
--- a/codemp/qcommon/RoffSystem.cpp
+++ b/codemp/qcommon/RoffSystem.cpp
@@ -647,7 +647,7 @@ qboolean CROFFSystem::Play( int entID, int id, qboolean doTranslation, qboolean 
 
 	roffing_ent->mEntID			= entID;
 	roffing_ent->mROFFID		= id;
-	roffing_ent->mNextROFFTime	= svs.time;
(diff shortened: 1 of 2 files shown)


ensiform:
Are either of these affected in SP as well?

ensiform (approved):
Looks good

Daggolin:
> Are either of these affected in SP as well?

No. For SP the relevant roff functions should be in module code and use `level.time`.
The `mVersion` type has been fixed for SP since 10dca1d9f3a3b9527ab19eac588f54275759bc32.

PR #1169, opened 2023-09-23
exclude builds that match debug+portable, and all of WinXP's debug and portable builds

Razish:
Just to speed up builds and reduce quota usage.
Also formatted `build.yml` with `yaml-language-server`.

Job count: 22 -> 14
Artifacts: 11 -> 9
Build time: 15-20m -> 13m (but most of that was waiting (for build agents?))

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,17 +5,17 @@ on:
   push:
     branches: [master]
     paths-ignore:
-      - '**.md'
-      - '*.txt'
-      - '.gitignore'
-      - 'docs/*'
+      - "**.md"
+      - "*.txt"
+      - ".gitignore"
+      - "docs/*"
   pull_request:
     branches: [master]
     paths-ignore:
-      - '**.md'
-      - '*.txt'
-      - '.gitignore'
-      - 'docs/*'
+      - "**.md"
+      - "*.txt"
+      - ".gitignore"
+      - "docs/*"
   release:
     types: [published]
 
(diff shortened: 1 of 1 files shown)


bartrpe (approved):
Reasonable change, I don't believe we need those things

.github/workflows/build.yml
```diff
@@ -28,6 +28,9 @@ jobs:
         arch: [x86, x86_64]
         build_type: [Debug, Release]
         portable: [Portable, Non-Portable]
+        exclude:
+          - build_type: Debug
+            portable: Portable
```
taysta, line 33: Is there any reason to do it differently here and on lines 84/85, here we are excluding, but there we just removed from matrix. I have seen users request a windows portable build in the past (mjt), to which I would direct them to the actions, Can we justify it's removal?

Razish: Lines 84/85 is removing _all_ {portable or debug} builds from WinXP, we only need to perform standard release builds to ensure it works. Anyone wanting a portable or debug build for WinXP are gonna have to do it themselves.

Here, we're only excluding the windows build if it's debug _and_ portable. So we'll get {release, portable} and {debug + non-portable} but not {debug, portable}.

So to be clear, we will provide portable Windows builds. Just not for WinXP.

taysta: Okay, that makes sense, thanks for clearing that up. In that case, LGTM.

SomaZ (approved):

PR #1170, opened 2023-09-25
[Shared] Add cl_mouseAspectScaling to correct UI mouse sensitivity for your resolution

Razish:
With this enabled, the cursor moves about the same rate as my desktop cursor.
Without this, horizontal movement is too sensitive on e.g. 1920x1080 windowed.

It's optional behind a cvar because without a breaking API change[1] or a more complex solution[2], small (sub-pixel) movements are completely discarded by being truncated.

[1] `int` -> `float` for `UI_MOUSE_EVENT` / `CG_MOUSE_EVENT`
[2] e.g. floating point accumulation buffer that emits movement events when it exceeds a whole number

### DIFF
--- a/code/client/cl_main.cpp
+++ b/code/client/cl_main.cpp
@@ -61,6 +61,7 @@ cvar_t	*cl_freelook;
 cvar_t	*cl_sensitivity;
 
 cvar_t	*cl_mouseAccel;
+cvar_t	*cl_mouseAspectScaling;
 cvar_t	*cl_showMouseRate;
 cvar_t	*cl_framerate;
 
@@ -1258,6 +1259,7 @@ void CL_Init( void ) {
 	cl_run = Cvar_Get ("cl_run", "1", CVAR_ARCHIVE_ND);
 	cl_sensitivity = Cvar_Get ("sensitivity", "5", CVAR_ARCHIVE);
 	cl_mouseAccel = Cvar_Get ("cl_mouseAccel", "0", CVAR_ARCHIVE_ND);
+	cl_mouseAspectScaling = Cvar_Get ("cl_mouseAspectScaling", "1", CVAR_ARCHIVE_ND);
 	cl_freelook = Cvar_Get( "cl_freelook", "1", CVAR_ARCHIVE_ND );
 
 	cl_showMouseRate = Cvar_Get ("cl_showmouserate", "0", 0);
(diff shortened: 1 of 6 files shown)


bartrpe (approved):
So I had it tested across various things:
- two merged 1920x1080 screens (essentially 3840x1080... I should one day get to try to list all the various bugs that happen at this ultra wide level)
- one 1920x1080
- tested every resolution set that's preset at r_mode options
- tested on 3840 x 2160 TV
- tested on old 1280x1024 screen

did that with various mouses too because why not

no bugs found

code/client/cl_main.cpp
```diff
@@ -1258,6 +1259,7 @@ void CL_Init( void ) {
 	cl_run = Cvar_Get ("cl_run", "1", CVAR_ARCHIVE_ND);
 	cl_sensitivity = Cvar_Get ("sensitivity", "5", CVAR_ARCHIVE);
 	cl_mouseAccel = Cvar_Get ("cl_mouseAccel", "0", CVAR_ARCHIVE_ND);
+	cl_mouseAspectScaling = Cvar_Get ("cl_mouseAspectScaling", "0", CVAR_ARCHIVE_ND);
```
SomaZ, line 1262: Since this is UI only, I'd actually prefer this to be on by default.

codemp/client/cl_main.cpp
```diff
@@ -2730,6 +2731,8 @@ void CL_Init( void ) {
 	// this should be set to the max rate value
 	cl_mouseAccelOffset = Cvar_Get( "cl_mouseAccelOffset", "5", CVAR_ARCHIVE_ND, "Mouse acceleration offset for style 1" );
 
+	cl_mouseAspectScaling = Cvar_Get( "cl_mouseAspectScaling", "0", CVAR_ARCHIVE_ND, "Scale mouse movement correctly in menus based on your desktop aspect ratio" );
```
SomaZ, line 2734: Same as above

code/client/cl_main.cpp
Daggolin: I would prefer to keep it off by default. At least from a modding perspective it's probably not nice if you switch between different clients for testing and the mouse acts different in one of them without a menu option to change the behavior.

codemp/client/cl_main.cpp
Daggolin: Same as above

Daggolin (approved):

code/client/cl_main.cpp
SomaZ: Hm, yea. Valid point I agree with.

taysta:
I think a QoL change like this should be defaulted on since the average user probably won't be aware of this cvar. 

Also, EternalJK and forks have had a similar change implemented and defaulted on for a long time, so the jarring difference in mouse scaling for these users would occur opposite of the situation described above.

I understand the arguments either way, as the base game has no such change and that could be considered as the "default" behaviour, so it's not a huge deal either way.

PR #1173, opened 2023-09-27
Add OpenJO build to GitHub actions

Razish:
This is based off the work done in #1169, which ~~should be merged first (then rebase this).~~ :white_check_mark: has been merged

Adds OpenJO build options to existing matrix (not a new variant), adds separate packaging + upload steps.
It's not added as a new matrix entry to reduce unnecessary jobs/work. A lot of the build process will be shared, and it's already built into different target dirs.

Normalised cmake configuration options across runners.

Fixed a couple incorrect context usages:

- `runner.build_type` should have been `matrix.build_type`, it was expanding to `""`.
- `runner.workspace` should probably be `github.workspace`: [github context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)

## TODO

- [x] Remove `UseInternalLibs`, it doesn't exist in OJK. May need to specify exactly which libs.
- [x] Can we parallelise `cmake --build` steps? We have [at least](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) 2 cores
- [x] Windows builds artifacts may be including a bunch of unnecessary libs since there's another parent folder (OpenJK/OpenJO)

## Future work

- [ ] Consider `OpenGL_GL_PREFERENCE=GLVND`? It is defaulting to legacy GL libs currently.


.github/workflows/build.yml
```diff
@@ -165,52 +172,61 @@ jobs:
...
 
-      - name: Create binary archive
-        if: ${{ matrix.build_type  == 'Release'}}
-        working-directory: ${{ runner.workspace }}/install/JediAcademy
+      - name: Create OpenJK binary archive
+        if: ${{ matrix.build_type == 'Release'}}
+        working-directory: ${{ github.workspace }}/install/JediAcademy
         shell: bash
         run: |
-          if [ ${{ matrix.arch }} == "x86" ]; then
```
Razish, line 202: The `chmod +x` doesn't appear to be necessary. It also doesn't apply to the SP binary.
We can reintroduce it if necessary

.github/workflows/build.yml
```diff
@@ -107,36 +112,38 @@ jobs:
...
-          
+
       - uses: actions/upload-artifact@v3
-        if: ${{ matrix.build_type == 'Release' }} 
+        if: ${{ matrix.build_type == 'Release' }}
         with:
-          name: OpenJK-windowsxp-${{ matrix.arch }}-${{ matrix.build_type }}-${{ matrix.portable }}
-          path: ${{ runner.workspace }}/build/bin
+          name: windowsxp-${{ matrix.arch }}-${{ matrix.build_type }}-${{ matrix.portable }}
+          path: ${{ github.workspace }}/build/bin
```
Razish, line 146: FIXME: reintroduce game prefix, maybe split upload step into two? why is this different to the other runners?

.github/workflows/build.yml
```diff
@@ -107,36 +119,44 @@ jobs:
...
-          
+        run: cmake --install . --config ${{ matrix.build_type }}
+
       - uses: actions/upload-artifact@v3
-        if: ${{ matrix.build_type == 'Release' }} 
+        if: ${{ matrix.build_type == 'Release' }}
         with:
           name: OpenJK-windowsxp-${{ matrix.arch }}-${{ matrix.build_type }}-${{ matrix.portable }}
-          path: ${{ runner.workspace }}/build/bin
+          path: ${{ github.workspace }}/build/bin/JediAcademy
```
taysta, line 152: This will fail on the release action as we upload the contents of the /JediAcademy/ folder here now, instead of the folder itself, but in the release action below we are still looking for the Jedi Academy folder when creating the zip in the create-release step.

.github/workflows/build.yml
```diff
@@ -282,11 +326,17 @@ jobs:
...
+          7z a -r OpenJK-windows-x86_64.zip ./OpenJK-windows-x86_64-Release-Non-Portable/JediAcademy/* '-x!msvcp*.*' '-x!vcruntime*.*' '-x!concrt*.*'
+          mv ./OpenJK-linux-x86-Release-Non-Portable/* OpenJK-linux-x86.tar.gz
+          mv ./OpenJK-linux-x86_64-Release-Non-Portable/* OpenJK-linux-x86_64.tar.gz
+          mv ./OpenJK-macos-x86_64-Release-Non-Portable/* OpenJK-macos-x86_64.tar.gz
+
+          7z a -r OpenJO-windows-x86.zip ./OpenJO-windows-x86-Release-Non-Portable/JediOutcast/* '-x!msvcp*.*' '-x!vcruntime*.*' '-x!concrt*.*'
+          7z a -r OpenJO-windows-x86_64.zip ./OpenJO-windows-x86_64-Release-Non-Portable/JediOutcast/* '-x!msvcp*.*' '-x!vcruntime*.*' '-x!concrt*.*'
+          mv ./OpenJO-linux-x86-Release-Non-Portable/* OpenJO-linux-x86.tar.gz
+          mv ./OpenJO-linux-x86_64-Release-Non-Portable/* OpenJO-linux-x86_64.tar.gz
+          mv ./OpenJO-macos-x86_64-Release-Non-Portable/* OpenJO-macos-x86_64.tar.gz
```
taysta, line 339: /JediAcademy/ and /JediOutcast/ need to be removed from lines 329, 330, 335, 336 for create-release to pass, due to changes from this PR outline in my earlier review comment in this file.

taysta (changes requested):

taysta (approved):
Looks good, hopefully this gets merged soon so people have better access to the OpenJO builds

SomaZ (approved):

PR #1177, opened 2023-10-08
[Shared] add a custom CMake target to copy compilation DB from build dir to source dir

Razish:
Configure with `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` to generate the compilation DB.
This PR will additionally copy this from the build directory to the source directory.
For example, this will provide better VSCodium support when using clangd instead of MS' C++ extension.

see links for further details:

- <https://clang.llvm.org/docs/JSONCompilationDatabase.html>
- <https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -182,6 +182,7 @@ pip-log.txt
 ## clangd
 #############
 .clangd
+compile_commands.json
 
 #############
 ## CLion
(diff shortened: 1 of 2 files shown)


taysta:
Is this missing compile_commands.json?

Razish:
No, that's a generated file that should be gitignored (which it is here).
It contains commands, directories etc that are only relevant for your machine/configuration.

taysta (approved):

SomaZ (approved):

PR #1182, opened 2023-11-02
Reintroduce QVM support

Daggolin:
# Prologue
The original jka release used native binaries, but the vanilla jamp and jampded engine binaries were still capable of loading QVM mods and both, the 2003 SDK and the 2013 release of the code could be used to compile QVMs for `jampgame`, `cgame` and `ui`. It is unknown whether anyone tried to make use of that before 2013, but there seems to be no publicly known mods that made use of QVMs, thus OpenJK removed support for QVMs early on. However, the interest in QVM mods has increased over time for better mod-compatibility accross different platforms and there is currently at least one server running jampded with a QVM based mod. For those reasons this pull request readds QVM support by porting the QVM code from jk2mv (which ported parts of it from ioq3) and adjusting the module APIs to support QVMs.

## Issues when using QVMs on vanilla jamp/jampded
- The string pointers written into the `glconfig_t` instances inside of `cgame` and `ui` would point to engine memory, making them invalid
- The notetrack callback for the roff system would pass a pointer to engine memory into the module, making it invalid (this issue already existed in jk2, but vanilla jk2 modules didn't properly support roff, leading to a very low chance of anyone actually having encountered the bug on there)
- A few of the NAV VM_Calls would pass `vec3_t` (pointer to float) into the game module, possibly affecting NPC pathfinding

# Pull Request
## QVM related changes
- Ports `vm_arm.cpp`, `vm_interpreted.cpp`, `vm_local.h` and `vm_x86.c` from mvdevs/jk2mv (which in turn got some of its code from ioq3 and multiple contributors)
- Ports various QVM related functions from mvdevs/jk2mv
- Reintroduces the cvars `vm_game`, `vm_cgame` and `vm_ui`
- Reintroduces the commands `vminfo` and `vmprofile`
- Maps G2 pointers to integer handles for the module APIs when using QVMs, because the types within the QVM are 32 bit, making 64 bit pointers incompatible
- Replaces `sharedEntity_t` in the engine with `sharedEntityMapper_t`. As the QVM types are 32 bit the `sharedEntity_t` struct doesn't match when running the engine as 64 bit build. To work around this the engine stores the correct addresses for the members of `sharedEntity_t` in `sharedEntityMapper_t` when the game module passes the shared entities into `SV_LocateGameData`.
- Solves the vanilla jamp/jampded issues by allocating additional memory for the QVM on load. This additional memory is used to write data into and then pass a pointer to that data into the module, which is valid, because it's within the datamask for the module.
- Replaces the exceptions in `qcommon.cpp` with `longjmp`, because throwing exceptions within a VM_Call was unreliable

## Other changes
- Fixes undefined behavior in `SV_ClipMoveToEntities`, which becmae obvious when replacing `sharedEntity_t` with `sharedEntityMapper_t`
- The `vm_legacy` cvar is no longer limited to `_DEBUG` builds to make potenital API tests with users easier, because it doesn't require extra builds
- Adds `com_dedicatedForceErrorsToFatal` cvar to be able to disable the `ERR_DROP` to `ERR_FATAL` rewrite for Com_Error calls on dedicated servers
- Blocks additional file extensions for write attempts from VMs (`dll`, `exe`, `bat`, `cmd`, `dylib`, `so`, `qvm`, `pk3`).
- Resolves filenames for the extension check to cover symlinks and potential bypasses of the checks
- Blocks a few characters in filenames specified by VMs (`<>:"|?*`)


mrwonko (commented):
I didn't get very far into the review before I got nerdsniped into trying to find a cleaner solution for the cross-module error throws/longjumps. But who knows when I'll get back to this, so here's what I have so far.

codemp/client/cl_uiapi.cpp
```diff
@@ -36,6 +39,62 @@ extern botlib_export_t *botlib_export;
 static uiExport_t *uie; // ui export table
 static vm_t *uivm; // ui vm, valid for legacy and new api
 
+typedef std::unordered_map<g2handle_t, CGhoul2Info_v*> g2HandleToG2_m;
+
+static g2HandleToG2_m g2Mapping;
+static g2handle_t g2NextHandle = (g2handle_t)1; // Start at 1, because 0 has special meaning
+
+static inline CGhoul2Info_v *CL_UI_G2Map_GetG2FromHandle( g2handleptr_t g2h )
```
mrwonko, line 47: Instead of duplicating all this g2mapping code, it would be nice to wrap it in a class so it can be instantiated twice

codemp/client/cl_cgameapi.cpp
```diff
@@ -356,6 +496,28 @@ static void CL_S_Shutup( qboolean shutup ) {
...
+	// The renderer should look up the handle on its own, but as we have to stay compatible with the modular renderer
+	// system we can't change the renderer for this and have to meet its expectations instead...
+
+	// Get a copy of the refEntity data. For a QVM mod the ghoul2 pointer at the end might contain some additional data,
+	// because the QVM stores it as 32 bit and if the engine is 64 bit it is going to read slightly out of bounds of the
+	// struct in the QVM, but CL_G2Map_GetG2FromHandle should take care of that by casting.
+	refEntity_t refEntCopy = *refEnt;
+
+	// Get the ghoul2 instance from the handle
+	refEntCopy.ghoul2 = CL_G2Map_GetG2FromHandle( (g2handleptr_t)refEnt->ghoul2 );
```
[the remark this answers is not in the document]
mrwonko: So I checked, and it looks like int is of equal size in both cases. But it's not guaranteed to be, so it would be nice to use int32_t instead. And there should be an assertion for the sizeof refEnt_t to be safe.

codemp/game/g_public.h
```diff
@@ -254,6 +254,86 @@ typedef struct sharedEntity_s {
...
+	int				noWaypointTime;		//Debouncer - so don't keep checking every waypoint in existance every frame that you can't find one
+	int				combatPoint;
+	int				failedWaypoints[MAX_FAILED_NODES];
+	int				failedWaypointCheckTime;
+
+	int				next_roff_time; //rww - npc's need to know when they're getting roff'd
+} sharedEntity_qvm_t;
+
+typedef struct sharedEntityMapper_s {
+	entityState_t	*s;				// communicated by server to clients
```
mrwonko, line 296: if this used references (or some kind of dumb smart pointer?) instead of raw pointers, we could avoid having to turn all the `.` into `->`, which would make this more appealing to merge downstream. But I've not looked at enough of the context to know if that's possible; particularly whether references' initialize-once-and-never-reassign semantics work here.

codemp/qcommon/common.cpp
```diff
@@ -289,7 +293,7 @@ void NORETURN QDECL Com_Error( int code, const char *fmt, ... ) {
 	}
 
 	if ( code == ERR_DISCONNECT || code == ERR_SERVERDISCONNECT || code == ERR_DROP || code == ERR_NEED_CD ) {
-		throw code;
+		longjmp(abortframe, code+1); // +1 to avoid 0 value
```
mrwonko, line 296: I think using longjmp instead of exceptions should probably be a separate PR. I have many thoughts.

codemp/qcommon/files.cpp
```diff
@@ -312,6 +312,77 @@ FILE*		missingFiles = NULL;
...
 #endif
 
+const char *get_filename_ext(const char *filename) {
+	const char *dot = strrchr(filename, '.');
+	if (!dot || dot == filename) return "";
+	return dot + 1;
+}
+
+// We don't want VMs to be able to access the following file extensions
+static char *invalidExtensions[] = {
```
mrwonko, line 322: an `std::array<const char*>` is nicer to work with; but this could arguably also be an `std::set` (though for a small set like this one, an array is likely more efficient)

codemp/qcommon/files.cpp
```diff
@@ -312,6 +312,77 @@ FILE*		missingFiles = NULL;
...
+qboolean FS_IsInvalidExtension( const char *ext ) {
+	int i;
+
+	// Compare against the list of forbidden extensions
+	for ( i = 0; i < invalidExtensionsAmount; i++ ) {
+		if ( !Q_stricmp(ext, invalidExtensions[i]) )
+			return qtrue;
+	}
+
+	// Additional check we don't currently need, but if support for another platform is added and the list above is not
```
mrwonko, line 343: I'd have done a compile-time-check using a `static_assert` to verify that `DLL_EXT` is included in `invalidExtensions`, but that might require more constexpr than C++11 gives us

codemp/qcommon/files.cpp
```diff
@@ -312,6 +312,77 @@ FILE*		missingFiles = NULL;
...
+	return qfalse;
+}
+
+// Invalid characters. Originally intended to be OS-specific, but considering that VMs run on different systems it's
+// probably not a bad idea to share the same behavior on all systems.
+qboolean FS_ContainsInvalidCharacters( const char *filename ) {
+	static char *invalidCharacters = "<>:\"|?*";
+	char *ptr = invalidCharacters;
+
+	while ( *ptr ) {
```
mrwonko, line 357: This is premature optimization, but I don't like this quadratic runtime. A single loop over filename with a `switch` for all invalid characters would be more efficient. Or you could use `std::find_first_of` for simplicity - the compiler will probably make that at least as efficient as the manual switch statement.

codemp/qcommon/files.cpp
```diff
@@ -312,6 +312,77 @@ FILE*		missingFiles = NULL;
...
+	// Resolve paths
+	resolved = Sys_ResolvePath( ospath );
+	if ( !strlen(resolved) ) return qtrue;
+	realPath = Sys_RealPath( resolved );
+	if ( !strlen(realPath) ) return qtrue;
+
+	// Check all three, the original input, the resolved and the real path, in case there is a symlink
+	// NOTE: Checking the ospath shouldn't be required, because we resolved the path, but let's check it anyway.
+	if ( FS_IsInvalidExtension(get_filename_ext(ospath)) || FS_IsInvalidExtension(get_filename_ext(resolved)) || FS_IsInvalidExtension(get_filename_ext(realPath)) ) {
+		Com_Printf( "FS_IsInvalidWriteOSPath: blocked writing binary file \"%s\" (%s) [%s].\n", ospath, resolved, realPath );
```
mrwonko, line 379: This feels like the wrong place for this log. This function should be pure (FS operations like following symlinks aside), and this log should be done by its caller.

mrwonko (commented):
gee, this is a large PR... I'll need to return to it and continue later.

codemp/game/g_public.h
mrwonko: Actually, I don't think you can overload `operator.`, so references are probably the only way to avoid adding another level of pointer indirection.

codemp/client/cl_cgameapi.cpp
```diff
@@ -37,6 +40,62 @@ extern botlib_export_t *botlib_export;
...
+}
+
+static void CL_G2Map_Update( g2handleptr_t *g2h, CGhoul2Info_v *g2Ptr )
+{ // Inserts and/or erases to/from the map and updates the handle pointer
+	if ( cgvm->dllHandle ) return;
+
+	g2handle_t *g2handle = (g2handle_t*)g2h;
+	if ( !*g2handle && g2Ptr )
+	{ // Got a 0 handle, but a non-0 pointer: add to map and set handle
+		// Unlikely to happen, but should we ever cycle through the whole integer range start searching for gaps
```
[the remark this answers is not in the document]
mrwonko: It might be more efficient to maintain a free list, i.e. keep track of released indices. It depends on the exact usage pattern, though.

codemp/qcommon/common.cpp
mrwonko: I did not understand that exceptions cause crashes in QVM JIT. That makes sense, and explains why its necessary to revert to the longjmp used in JK2 and Q3. I hate that this prevents the use of custom destructors in any function that can (directly or indirectly) call `Com_Error`, but translating between longjmps and exceptions at the module boundary every time does not seem feasible.

- [ ] I should annotate all functions that are affected by this to see the impact and check if it's safe. I'd kind of like those annotations to be included in this PR, but I suspect there will be too many of them, leading to many potential merge conflicts downstream.

codemp/client/cl_cgameapi.cpp
```diff
@@ -213,9 +272,26 @@ qboolean CGVM_NoUseableForce( void ) {
...
 
+static size_t vec3_size = sizeof(vec3_t);
 void CGVM_GetOrigin( int entID, vec3_t out ) {
 	if ( cgvm->isLegacy ) {
-		VM_Call( cgvm, CG_GET_ORIGIN, entID, reinterpret_cast< intptr_t >( out ) );
+		float *out_ptr;
+		float *out_ptr_vm;
+
+		if ( !cgvm->dllHandle ) {
+			out_ptr = (float*)VM_ExtraMemory_ClaimData(cgvm, (float*)out, vec3_size);
```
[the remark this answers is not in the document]
mrwonko: And since it looks like the longjmp is mandatory, RAII is not an option here :(

codemp/client/cl_cgameapi.cpp
```diff
@@ -37,6 +40,62 @@ extern botlib_export_t *botlib_export;
...
+static inline CGhoul2Info_v **CL_G2Map_GetG2PtrFromHandle( g2handleptr_t *g2h )
+{ // Returns a pointer to the g2 object pointer in the map so g2 functions can update the pointer
+	// Native libraries should not use the pointer, but in theory they could use
+	// it. Thus we don't perform any mapping for native libraries to avoid
+	// issues with custom modules.
+	if ( cgvm->dllHandle ) return (CGhoul2Info_v **)g2h;
+
+	g2handle_t g2handle = *((g2handle_t*)g2h);
+	if ( !g2handle )
+	{ // Special case: the g2 handle is not valid, yet. Return a pointer to a static temporary pointer. Insertion is handled by calling CL_G2Map_Update after calling the G2API
```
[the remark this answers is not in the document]
mrwonko: I understand this better now, and it seems fairly easy to get rid of this hack: 1fd2945c183d14d54962d734425c436bb9968a95

codemp/server/server.h
```diff
@@ -152,6 +154,7 @@ typedef struct client_s {
 	int				lastClientCommand;	// reliable client message sequence
 	char			lastClientCommandString[MAX_STRING_CHARS];
 	sharedEntity_t	*gentity;			// SV_GentityNum(clientnum)
```
mrwonko, line 156: The entity this points to can be either in the QVM or Native, right? So there's no way to use it safely, I think this should be deleted, only `gentityMapper` can be used safe.

mrwonko: fix in f81e89390cae67f7cb8ad33d98048919f6604e7f

entdark:
When the merge @Daggolin?!

PR #1183, opened 2023-11-06
Unix: Fix argv[0] getting overwritten on launch

Daggolin:
Calling `Sys_Dirname` on `argv[0]` lead to slashes in `argv[0]` getting overwritten on Linux (and potentially other non-Windows platforms), causing some external script to fail sending signals to the right process, because the path changed.

This pull request adjusts `Sys_Dirname` and `Sys_Basename` to work with a copy of the given string instead of the original. This makes the behavior of both functions more consistent between Windows and other platforms.

### DIFF
--- a/shared/sys/sys_public.h
+++ b/shared/sys/sys_public.h
@@ -141,8 +141,8 @@ char    *Sys_DefaultAppPath(void);
 #endif
 
 char	*Sys_DefaultHomePath(void);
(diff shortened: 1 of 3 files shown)


SomaZ (approved):

bartrpe (approved):
Simple and straightforward as hell

ensiform (approved):

PR #1185, opened 2023-11-07
FS Features

Daggolin:
This pull request aims to improve a few FS related issues and to provide new features.

1) [MP] The `fs_forceGame` cvar from jk2mv. The user can set the force game on startup to ensure configs, screenshots, etc. are written to and read from the specified folder. The server provided `fs_game` value is still used for reading, so mods in custom folders and pk3s downloaded from the server should still work.

2) [MP] Files downloaded from the server get prefixed with `dl_` in their name. Pk3s with this prefix are not loaded unless the server references them. If a server provides a pk3s that overrides all default textures, models, etc. it is only used on servers offering the same file for download. Playing on other servers the pk3 is ignored. This is ported form jk2mv as well.

3) [MP] Reflists from jk2mv. The reflists can be used to control file referencing when `sv_allowDownload` is enabled. For instance to only offer selected files (`ref_whitelist.txt`), block specific files (`ref_blocklist.txt`) and to offer files that wouldn't automatically be picked up by the download system (`ref_forcelist.txt`).

4) [Shared] Native libraries (dlls on Windows) are no longer automatically unpacked and used from pk3s, unless the game is started with `com_unpackLibraries` enabled.

5) [Shared] The network initialization is now performed before the filesystem initialization to avoid undesired side effects from network sockets having high fd values. For instance high kernel load on some systems due to high values in `select` calls.

6) [Shared] The `fs_restart` command from jk2mv. If no module has an active handle for a file within a pk3 the filesystem can be reloaded at runtime. This can be useful when testing/developing mods or when installing new maps, because intermediate map changes are no longer required to reload the filesystem.

7) [Shared] Increase pk3 limit. The hardcoded limit of 1024 files has been removed. The game tries to set the fd limit to 4096 on launch now. The start argument `-maxfds` can be used to override the default limit of 4096 (if the operating system permits the user to do so). This is ported from jk2mv as well.

The code ported from jk2mv was written by @ouned, @aufau, myself and possibly other contributors over time.


Daggolin:
Pushed additional changes:

8. [MP] Ask users to confirm downloads (cvar: cl_downloadPromt) and print infos about the current download without relying on the ui module (cvar: cl_downloadOverlay). The prompt is similar in function to the one in jk2mv, but jk2mv uses a custom ui module (mvmenu) and menu files during the downloads. To avoid relying on matching ui modules the window in this pull request is drawn internally within the engine.

9. [MP] Set default value of cl_allowDownload to 1, because cl_downloadPrompt defaults to 1 and thus a user is asked about every single download by default anyway.

Daggolin:
Pushed additional changes:

10. [MP] Skip download attempts if the server has `sv_allowDownload` set, but the value indicates it is disabled. This avoids showing the download prompt and trying to download files when the server doesn't even allow downloads.

11. [MP] Adjust CMakeFiles to add the OpenJK modules to the base folder for win32 builds. As unpacking of dlls is disabled by default the user may decide to install OpenJK into a folder without any unpacked vanilla modules. Thus OpenJK should provide modules for base on win32 now, too.

Razish (approved):

taysta (approved):

SomaZ (dismissed):

SomaZ (approved):

Razish (approved):

PR #1187, opened 2023-11-12
Fix r_showtris

xycaleth:
Depth test should be disabled when rendering wireframes so we can see the triangles being drawn behind the front-most surfaces.


Daggolin (approved):

xycaleth:
Already part of #1179

PR #1188, opened 2023-11-12
A few small changes (cl_filterGames + MAX_PATCH_PLANES increase)

Daggolin:
This pull request introduces a new cvar `cl_filterGames` and defaults it to `MBII MBIIOpenBeta`. Servers using an `fs_game` value that's on the list specified in `cl_filterGames` are going to be ignored on the serverlist. The default value seems to cover the majoriy of MBII servers.

Additionally the `MAX_PATCH_PLANES` value got increased from `2048` to `4096`, because some maps (at least for jk2) require a higher limit.

### DIFF
--- a/code/qcommon/cm_patch.h
+++ b/code/qcommon/cm_patch.h
@@ -62,7 +62,7 @@ properly.
 
 
 #define	MAX_FACETS			1024
-#define	MAX_PATCH_PLANES	2048
+#define	MAX_PATCH_PLANES	4096 // Was 2048 on vanilla jka
 
 typedef struct {
 	float	plane[4];
--- a/codemp/client/cl_main.cpp
+++ b/codemp/client/cl_main.cpp
@@ -105,6 +105,8 @@ cvar_t  *cl_lanForcePackets;
 
 cvar_t	*cl_drawRecording;
 
+cvar_t	*cl_filterGames;
+
 vec3_t cl_windVec;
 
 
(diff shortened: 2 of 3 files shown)


Daggolin:
> Seems a little racist to filter out an entire mod by default huh..

Good thing it's not hardcoded like it was on the fork you maintained. ^^

For anyone unfamiliar with the current situation: MBII is a total conversion for multiplayer. It requires new modules (game/cgame/ui), new pk3s (menus, maps, models, textures, ...) and possibly other things to be able to join a server. The mod can NOT be downloaded ingame (neither using the vanilla client nor OpenJK) and when you try to join the server you drop back to the main menu with an error message most players probably don't understand (for instance: `invalid entityState field count`). There is no helpful meassges printed, the player is left wondering what the issue is and why they can't play online or how to resolve it. While writing this message I checked the serverlist: there is currently 151 servers of which 56 match the default `cl_filterGames` proposed in this pull request. That means more than one third of the servers (almost 40 percent) are not usable with OpenJK, because they require a total conversion mod. This makes it hard for players to find actual servers they can join and in recent times players have already asked for help on how to fix their game, because they ran into error messages caused by trying to join MBII servers (with more than a third of servers requiring MBII that's not surprising). Who knows how many players tried to play online and gave up, not finding their way into community chatrooms to ask for help?

As I don't play MBII myself I may be wrong on this, but as far as I know MBII only supports the vanilla client and it requires a custom launcher to run the game. These defaults should not affect MBII.

> Really though, it shouldn't blacklist an fs_game if the client is currently using it.

Good catch, thank you. Added a check for that. If MBII supports OpenJK at some point just launching the game with the right fs_game should counteract the default value provided by `cl_filterGames` now.

Razish (approved):

SomaZ (approved):

Issue #1189, opened 2023-11-15
Rend2 out-of-bounds write while outside the map in the negative direction.


Charlese2:
Reporting a bug? Please make sure you've given the following information - thanks!

**Operating system and version:**
Windows 11 Version 22H2 (OS build 22621.2428)

**Is this for single player or multiplayer?**
Multiplayer

**Description of the bug (and if possible, steps to reproduce the bug):**
The Remainder of a negative Dividend with a positive Divisor can be a negative number. For example this modulo `-1 % 3` would result in `-1`. If `centerZoneOffsetX` (or Y) is a low enough number by noclipping outside the map while going in a negative direction, it will make the Dividend negative. 
https://github.com/JACoders/OpenJK/blob/4c422eaf45573d5b00047e84a9941e5c67e7c507/codemp/rd-rend2/tr_weather.cpp#L1144-L1149
If `chunkIndex` ends up as a negative number it writes outside of the `zoneOffsets` buffer.
https://github.com/JACoders/OpenJK/blob/4c422eaf45573d5b00047e84a9941e5c67e7c507/codemp/rd-rend2/tr_weather.cpp#L1134

**What did you expect to happen instead?**
I expected it to not crash from stack corruption from noclipping outside the map. I know it is a obscure use case.

Syrrr:
I can't seem to replicate this on e.g. t1_taspir, but I'm sure nothing would inherently break if we just checked that value and negated it?

Will do and open up a PR.

PR #1190, opened 2023-11-15
rend2 - fixed out-of-bounds crash with weather

Syrrr:
See issue #1189 

The game will crash, if a map features weather (e.g. vjun1) and you noclip far enough beneath the map.

This was caused by an out-of-bounds exception caused by chunkIndex sometimes being negative.

### DIFF
--- a/codemp/rd-rend2/tr_weather.cpp
+++ b/codemp/rd-rend2/tr_weather.cpp
@@ -1145,7 +1145,7 @@ void RB_SurfaceWeather( srfWeather_t *surf )
 				chunkIndex += (int(centerZoneOffsetY + numMinZonesY) + y + 1) % 3 * 3;
 
 				if (chunkIndex < 0) {
-					chunkIndex *= -1;
+					chunkIndex += 9;
 				}
 
 				VectorSet2(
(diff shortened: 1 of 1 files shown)


codemp/rd-rend2/tr_weather.cpp
```diff
@@ -1143,6 +1143,11 @@ void RB_SurfaceWeather( srfWeather_t *surf )
 			{
 				chunkIndex  = (int(centerZoneOffsetX + numMinZonesX) + x + 1) % 3;
 				chunkIndex += (int(centerZoneOffsetY + numMinZonesY) + y + 1) % 3 * 3;
+
+				if (chunkIndex < 0) {
+					chunkIndex *= -1;
```
Charlese2, line 1148: Because the `%` operator uses _truncated division_ just flipping the sign won't lead to a continuous remainder. 
[image ](https://en.wikipedia.org/wiki/Modulo)
With just flipping the sign, it causes clumping of the rain in the shader. 
image

Syrrr: Oh, yeah! You are right. I flew around and didn't spot that issue.

ensiform: I do not know if flipping the sign is actually the best fix all together either, but I could be wrong.

PR #1124, opened 2022-09-02
[Shared] provide option to use AddressSanitizer

Razish:
:shrug: why not

PR #1198, opened 2024-01-17
Added support ccache for speedup compilation on Linux

GermanAizek:
more information:
https://ccache.dev/
https://en.wikipedia.org/wiki/Ccache


Razish (approved):

taysta (approved):

Razish:
Just rethinking this. Should it be gated behind an `option(UseCCache)` like #1124?
What if i have `ccache` installed, but I don't want to use it for this build right now?
Bonus points for having the option exposed only if `ccache` is detected 😁

PR #1199, opened 2024-02-08
replace PATCH_STITCHING preprocessor define with r_patchStitching cvar

Razish:
Seems to work.

Trailing whitespace removal came along for the ride :shrug:


SomaZ:
I'm not sure whats the usecase of this, but it looks fine to me. Only thing I'd like to be done too is applying the same thing to SP because I think we want to reduce differences in the MP and SP renderers over time, not increase them.

SomaZ:
Ah, nvm me. SP doesnt have patch stiching implemented at all.

SomaZ (approved):

PR #1201, opened 2024-02-10
[Shared] misc loop overrun fixes

Razish:
These looked wrong :shrug:
I stumbled upon the `PW_NUM_POWERUPS` by accident, then found the rest with `<=\s*(\w+_)*NUM_\w+`

~I'm not 100% sure on the `NUM_BEAM_SEGS` loops in the renderers - @SomaZ?~ reverted!

### DIFF
--- a/code/rd-vanilla/tr_surface.cpp
+++ b/code/rd-vanilla/tr_surface.cpp
@@ -1111,7 +1111,7 @@ static void RB_SurfaceBeam( void )
 	}
 
 	qglBegin( GL_TRIANGLE_STRIP );
-	for ( i = 0; i <= NUM_BEAM_SEGS; i++ ) {
+	for ( i = 0; i < NUM_BEAM_SEGS; i++ ) {
 		qglVertex3fv( start_points[ i % NUM_BEAM_SEGS] );
 		qglVertex3fv( end_points[ i % NUM_BEAM_SEGS] );
 	}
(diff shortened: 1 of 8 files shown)


code/rd-vanilla/tr_surface.cpp
```diff
@@ -1111,7 +1111,7 @@ static void RB_SurfaceBeam( void )
 	}
 
 	qglBegin( GL_TRIANGLE_STRIP );
-	for ( i = 0; i <= NUM_BEAM_SEGS; i++ ) {
+	for ( i = 0; i < NUM_BEAM_SEGS; i++ ) {
```
kungfooman, line 1114: Didn't test anything, but `<=` looks intentional to close it off, why else would there be a modulo? Did you test these changes?

SomaZ: > Didn't test anything, but `<=` looks intentional to close it off, why else would there be a modulo? Did you test these changes?

^ This is correct. Its used intentionally for sealing.

Razish: Thanks for confirming, that makes sense. I was having trouble visualising what vertices were being emitted, and it's hard to test ingame since nothing uses them and it doesn't seem to render anything anyway :shrug:

SomaZ (approved):

PR #1202, opened 2024-02-10
[Meta] trim EOL whitespace

Razish:
Not sure if we want to merge this right now, but this is what it looks like to normalise/trim our EOL whitespace.

I ran a search/replace of `[ \t]+\n` -> `\n` in paths matching: `.c,.cpp,.h,.glsl,.rc,.txt,.cfg,.cmake` excluding `lib/`

## Future work

In a perfect world, I'd introduce clang-format with a CI step to fail if it doesn't adhere. But that would wreak havoc on forks now. It might be too late.

- https://github.com/JACoders/OpenJK/pull/1203


codeJK2/cgame/cg_effects.cpp
```diff
@@ -708,7 +708,7 @@ static void CG_CalcBiLerp( vec3_t verts[4], vec3_t subVerts[4], vec2_t uv[4] )
 	VectorMA( temp,			uv[3][1],			subVerts[3], subVerts[3] );
 }
 // bilinear
-//f(p',q') = (1 - y) � {[(1 - x) � f(p,q)] + [x � f(p,q+1)]} + y � {[(1 - x) � f(p+1,q)] + [x � f(p+1,q+1)]}. 
+//f(p',q') = (1 - y) � {[(1 - x) � f(p,q)] + [x � f(p,q+1)]} + y � {[(1 - x) � f(p+1,q)] + [x � f
```
SomaZ, line 711: This doesn't look right I think :)

SomaZ (dismissed):

PR #1203, opened 2024-02-10
POC: Introduce clang-format

Razish:
# DO NOT MERGE

This is a proof of concept for what introducing `clang-format` would look like.

Highly destructive for forks, but _so_ much nicer to work with.
I typically work with format-on-save, but have provided a helper script (`scripts/format-code.sh`) that could also be repurposed by CI to check if PRs adhere to code formatting rules.

Ref #1202


bartrpe:
uh, incorporating that into the forks would be a nightmare beyond belief

I'm afraid at this point it kinda has to stay the way it is. I'm afraid OpenJK reached the point in development where doing full reformatting is just not viable anymore ;_;

Razish:
I wonder if it would be easier for forks if they pre-formatted with the same `clang-format` config before merging.
Would still require large coordination and likely still be an extremely messy merge, but...probably smaller blast radius?
Still not worth it :upside_down_face:

PR #1204, opened 2024-02-10
[SP] cap the timescale sensitivity scale when timescale > 1.0

Razish:
If you set timescale > 1.0 in SP, your mouse sensitivity gets scaled up with it.

It looks like the intention was to slow your mouse sensitivity down when using force speed/rage, so let's cap timescale's influence to < 1.0


bartrpe (approved):
- code is correct (there's barely any actual code change, just value change anyway lol)
- i didn't find any other place that would affect these lines running through it with Visual Studio debugger
- it builds
- it works correctly ingame

SomaZ (approved):
Since this is just SP I consider this a QoL improvement.

Daggolin (approved):
Considering force speed already got a special case it might be more intuitive to completely unlink sensitivity from timescale or to add a cvar to enable/disable the scaling for timescale.

Either way, I am approving this, because it feels more consistent than what we currently have.

PR #1212, opened 2024-02-20
[MP] fix water jumping discrepancy

Razish:
reverts a change introduced in Razish/jasdk_modbase@09edabf but provide an alternative fix that better matches SP code/behaviour with as few changes as possible
this lets you jump out of the water pools in `taspir2` like you can in SP
introduces `g_fixWaterJump` (default: 0) / `LEGACYFIX_WATERJUMP`

discovered by @CansecoDev

SP:
https://github.com/JACoders/OpenJK/assets/844370/5ebdd10e-7ca5-4ab9-bf3e-c581b5934c25

MP original behaviour:
https://github.com/JACoders/OpenJK/assets/844370/2423b20b-52ae-4ab3-ba95-ad38ccff5e6d


SomaZ (approved):

Daggolin:
I don't like that the `CS_LEGACY_FIXES` configstring is evaluated manually in every place that relies on it. It would be a lot cleaner if the value were passed into a new member of the `pmove_t` struct. That way the pmove code wouldn't need as many game/cgame checks and cgame could parse the configstring once instead of doing it in multiple places.

I am also not sure about the name of the cvar. If I recall correctly most players refer to the bug that allows you to jump back to your maximum height when leaving the water as water jump. As far as I can tell this fix is unrelated to that bug.

PR #1216, opened 2024-02-26
[Shared] add con_timestamps 2 to disable timestamps in the notify lines only

Razish:


code/client/cl_console.cpp
```diff
@@ -631,8 +632,11 @@ void Con_DrawNotify (void)
 		if (time > con_notifytime->value*1000)
 			continue;
 		text = con.text + (i % con.totallines)*con.rowwidth;
-		if (!con_timestamps->integer)
+		if (con_timestamps->integer == 0 || con_timestamps->integer == 2) {
+			// don't show timestamps in the notify lines
 			text += CON_TIMESTAMP_LEN;
+			lineLimit -= CON_TIMESTAMP_LEN;
```
Razish, line 638: Without this, if there are multiple notify lines it will start printing the timestamp from line 2 at the end of line 1 :facepalm:

bartrpe (approved):
Simplest code ever, it's so simple even I can say everything is correct, there's nothing that could go wrong with these changes.

SomaZ (approved):

Daggolin (approved):

PR #1217, opened 2024-02-27
[MP] fix g_weaponDisable + g_forcePowerDisable regression when creating listen servers

### LINKED ISSUE #1215: Enabling lightsaber only and disable force in multiplayer doesn't work

Windows 11 23H2 build 22631.3155 x64

Multiplayer

Enabling lightsaber only and disable force in a multiplayer solo match doesn't work, the game still starts with force and weapons enabled.
Open a multiplayer game, enable Lightsaber only: Yes and Disable Force: Yes
Watch as bots spawn with weapons and force!
This feature works in retail, in openJK it's somehow broken.

The game starts without weapons and force with the feature tuned off! In retail it works as expected.

EDIT: Better title and edited post.

Razish:
re-enabling this block that turns e.g. `g_weaponDisable 1` --> `g_weaponDisable 524279`
also fix some stray uninitialised warnings.

fixes #1215

### DIFF
--- a/codemp/ui/ui_main.c
+++ b/codemp/ui/ui_main.c
@@ -7437,7 +7437,6 @@ static void UI_BuildServerDisplayList(int force) {
 	lanSource = UI_SourceForLAN();
 
 	if (force) {
-		// numinvisible = 0;
 		// clear number of displayed servers
 		uiInfo.serverStatus.numDisplayServers = 0;
 		uiInfo.serverStatus.numPlayersOnServers = 0;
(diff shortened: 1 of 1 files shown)


codemp/ui/ui_main.c
```diff
@@ -7440,7 +7437,7 @@ static void UI_BuildServerDisplayList(int force) {
 	lanSource = UI_SourceForLAN();
 
 	if (force) {
-		numinvisible = 0;
+		// numinvisible = 0;
```
mrwonko, line 7440: isn't checking in commented-out code a bad practice, generally?

Razish: Yeah, but removing some of these lines would have left empty if blocks followed by non-empty else-if blocks. In that case I'd rather there be some context as to why the control flow is how it is, which an empty block does not convey

Razish: Removed now.

ensiform (approved):

SomaZ (approved):

PR #1220, opened 2024-02-29
[Shared] Add exit command as an alias to quit

Razish:
I can't tell you how many times I've tried to close a server with `exit` like most other CLI programs, interpreters, shells etc.

Quality of life :sparkles:

### DIFF
--- a/code/qcommon/common.cpp
+++ b/code/qcommon/common.cpp
@@ -1124,6 +1124,7 @@ void Com_Init( char *commandLine ) {
 		// init commands and vars
 		//
 		Cmd_AddCommand ("quit", Com_Quit_f);
+		Cmd_AddCommand ("exit", Com_Quit_f);
 		Cmd_AddCommand ("writeconfig", Com_WriteConfig_f );
 
 		com_developer = Cvar_Get ("developer", "0", CVAR_TEMP );
--- a/codemp/qcommon/common.cpp
+++ b/codemp/qcommon/common.cpp
@@ -1207,6 +1207,7 @@ void Com_Init( char *commandLine ) {
 			Cmd_AddCommand ("freeze", Com_Freeze_f);
 		}
 		Cmd_AddCommand ("quit", Com_Quit_f, "Quits the game" );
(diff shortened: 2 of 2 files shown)


Razish:
Note there's a slight QoL decrease because `ex<TAB>` no longer completes to `exec` :thinking:
I'll leave this to the masses to decide which is better.

bartrpe (approved):
"exec" is a command used, in general, by slightly more experienced players so not having it on full auto-completion is less of a hindrance than adding the common "console-killer command" imo.

Razish:
We could also argue that `ex<TAB>` for `exec` is only saving 1 keystroke (`c`). Not likely to end the world.

taysta:
> We could also argue that `ex<TAB>` for `exec` is only saving 1 keystroke (`c`). Not likely to end the world.

Maybe over the top but in my fork I added a cvar for the exit command, default to on. I advised the few who complained about it interrupting their exec muscle memory to disable the exit command with the cvar.

Razish:
I guess that works, but it's not worth the additional complexity to me.
We go all-in or we don't add it at all IMO.

SomaZ (approved):
I'm okay with that change. I doubt anyone will complain. If someone has a valid reason against this, we can easily revert this again.

xiaoxiao921:
I just write `qui gon jinn` in the console and it works

ensiform:
> I just write `qui gon jinn` in the console and it works

Unsupported in openjk 😂

Daggolin:
I am rather cautious regarding new built-in commands on OpenJK. OpenJK has already added a built in `help` command that overrides the `help` commands from custom serverside mods, because the client no longer forwards the command to the server (unless the client uses `cmd help`). And EternalJK added an internal `login` command that no longer forwarded the command on servers that were using serverside mods EternalJK was not familiar with. So I don't really like the idea of adding `exit` as alias for `quit`.

I am with @taysta on this: I would prefer to have a cvar to enable this (and I would prefer it defaulting to off). However at that point one could consider to add a generic alias system to let users define custom rewrites (which could also be useful when playing on custom mods).

PR #1223, opened 2024-03-03
[Shared] fix UI transition/fade effects being framerate dependent

Razish:
Currently the `transition`/`fade` keywords in menus will behave differently according to your FPS.
This PR makes it consistent with the original behaviour as if you had 60 FPS - which is likely what they were designed for.

You can see these in effect in SP from the main menu:

- `New` (transition)
- `Controls` (fade in)

Demonstration of before/after this fix from JA++ with a custom menu:
https://github.com/JACoders/OpenJK/assets/844370/e84b79f1-5da6-4f6a-99fc-f67782263605

Demonstration of fixed behaviour in existing SP menu:
https://github.com/JACoders/OpenJK/assets/844370/16814f7c-94f3-4df3-9631-275bccd461ba

### DIFF
--- a/code/ui/ui_shared.cpp
+++ b/code/ui/ui_shared.cpp
@@ -7788,6 +7788,7 @@ static qboolean Item_Paint(itemDef_t *item, qboolean bDraw)
 	int		xPos,textWidth;
 	vec4_t red;
 	menuDef_t *parent;
+	const float tAdjust = DC->frameTime / (1000.0 / UI_TARGET_FPS);
 	red[0] = red[3] = 1;
 	red[1] = red[2] = 0;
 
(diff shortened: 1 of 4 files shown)


code/ui/ui_shared.cpp
```diff
@@ -7812,17 +7813,17 @@ static qboolean Item_Paint(itemDef_t *item, qboolean bDraw)
 		if (DC->realTime > item->window.nextTime)
 		{
 			float rx, ry, a, c, s, w, h;
-			item->window.nextTime = DC->realTime + item->window.offsetTime;
+			item->window.nextTime = DC->realTime + (item->window.offsetTime * tAdjust);
```
Razish, line 7816: I'm not sure if also adjusting the debouncer is correct since we're adjusting the translation based on the fractional current frametime / 60fps frametime.

bartrpe (approved):
I'm not the one to give a solid opinion on the good coding practice and whether solution is proper for the environment etc. buuut

I've tested the entire UI, checked the references and whether the adjustment is provided in every required variable & function and I can say it is. Checked the references to the timer, ran through call stack and reference tree, everything looks fine, every place that would need the adjustment, has it.

ensiform (approved):
I don't know for sure either but does MP have the correct frametime?

SomaZ (approved):

code/ui/ui_shared.cpp
SomaZ: I think you can be sure that most users have exactly 60 fps (vsync with most common displays) or more in the menus, thus this should only decrease the amount of tiggered updates. One could argue that this is bad for higher freqency displays, as the interpolation won't work in this case for the frames in between. Probably better to get rid of it, but I wouldn't mind if menu anmiations are capped at the defined fps.

Issue #1226, opened 2024-04-05
Oversized fonts in game menu


Blank99:
**Operating system and version:** Windows 11, Using custom resolution of 1920x1080.
**Is this for single player or multiplayer?** SP
**Description of the bug (and if possible, steps to reproduce the bug):** the ingame menu's font describing items/weapons/force powers is too large and spills over
![image](https://github.com/JACoders/OpenJK/assets/13709902/05396038-a444-483b-9fcb-9f8bb41247d5)
EDIT: Fixed the image

**What did you expect to happen instead?** Regular font size

ensiform:
Did you install any mods?

Blank99:
Not at that time, tried GGFontsHQ mod and it didn't fix it. Tried JKA Font Mod Typomoderno and it did, ugly, but it does the work. Sadly, only font mod that seems to change that font so far out of a few I've tested.

Blank99:
Checking again... I think it's using the wrong font? If I run the regular exe of the game I get in that screen a font that is not allcaps instead of the big fat menu font.

ensiform:
> Checking again... I think it's using the wrong font? If I run the regular exe of the game I get in that screen a font that is not allcaps instead of the big fat menu font.
> 

Yes I believe it's related to a change by @Daggolin but it may be difficult to track down due to the poorly implemented menu code in SP.

Thanks for the find.

PR #1227, opened 2024-04-07
[SP] Font fixes

Daggolin:
This pull request fixes 2 bugs in SP:
 - the datapad showed incorrect fonts for the weapons and force tabs (#1226)
 - the r_reloadfonts command made all fonts vanish

Additionally it adjusts `MenuFontToReal` in the ui module to be more flexible in case of menu files that don't actually load the default fonts.

### DIFF
--- a/code/cgame/cg_local.h
+++ b/code/cgame/cg_local.h
@@ -1231,4 +1231,6 @@ void	CG_ClearLightStyles( void );
 void	CG_RunLightStyles( void );
 void	CG_SetLightstyle( int i );
 
+int CG_MagicFontToReal( int menuFontIndex );
+
 #endif	//__CG_LOCAL_H__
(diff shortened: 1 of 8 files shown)


Razish (approved):
What a mess that font code is, thanks for dealing with it.

SomaZ (approved):

Daggolin:
> What a mess that font code is, thanks for dealing with it.

Well, I was the one who ported the JK2MV font variant code to OpenJK and thought: for consistency I should add it to MP and SP. So I am the one who stepped into that trap of horrible SP font handling. Had I realized there is even magic numbers in the cgame module of SP I might have left out the SP support to avoid corner cases like this and just put a warning comment into the pull request. Now I am trying to fix what I broke.

PR #1238, opened 2024-08-18
Update built in SDL version for windows

ensiform:
to 2.30.6

Make r_sdlDriver settable at startup to choose which SDL backend is used "wayland" for example.

Add r_allowScreensaver cvar to allow enabling screen to sleep while game is running.

Turn on hint to bring back pre-SDL 2.0.14 behavior for alt-tab on focus loss.

Add s_sdlDriver to choose which backend is used for SDL audio ("pulseaudio" or "pipewire" for example)

Silence Apple OpenGL deprecated warnings


taysta (approved):
Can confirm the sdl_*.cpp changes have worked fine since merging them into TaystJK in my latest SDL update.

PR #1243, opened 2024-10-18
Optimized ui_syscalls with null checks and cleanup

Nexxus67:
Refactored ui_syscalls.cpp to include null pointer checks and removed unused syscall comments for clarity. Improved code safety by validating input values, ensuring stable execution.


ensiform:
Most of these are unnecessary. And null setcolor rgba is explicitly allowed in the API so this is an invalid check.

PR #1247, opened 2024-11-19
Support JKA models in OpenJO 

taysta:
Quickly ported from https://github.com/mvdevs/jk2mv/commit/ef336f9fcef8ddb9ec5f642166d5d3727ff73bea, https://github.com/mvdevs/jk2mv/commit/3cd7e5313fb7a8ebb5b6cfb2c236987c33298412, https://github.com/mvdevs/jk2mv/commit/14cb6c9cddc6937c9f7ce00123c300086dacf944
Tested in-game, appears to work as expected.

### DIFF
--- a/code/rd-vanilla/tr_ghoul2.cpp
+++ b/code/rd-vanilla/tr_ghoul2.cpp
@@ -3659,8 +3659,7 @@ qboolean R_LoadMDXM( model_t *mod, void *buffer, const char *mod_name, qboolean 
 	{
 		isAnOldModelFile = true;
 	}
-#endif
-#ifdef JK2_MODE
+#else
 	bool isANewModelFile = false;
 	if (mdxm->numBones == 53 && strstr(mdxm->animName, "_humanoid"))
 	{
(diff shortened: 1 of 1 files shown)


Razish (dismissed):

Daggolin (approved):
The code is exactly like on jk2mv and a quick test shows it's working. I personally would've preferred to have the additional ifdef as else, because they're mutually exclusive, depending on the target (openjk sp vs. openjo sp), but I am approving this so it can be merged.

Current:
```c
#ifndef JK2_MODE
...
#endif
#ifdef JK2_MODE
...
#endif
```

Peferred:
```c
#ifndef JK2_MODE
...
#else
...
#endif
```

When this is merged @SomaZ might have to adapt it for #1179 to keep feature parity for rd-vanilla and rd-rend2.

taysta:
I'll make that adjustment when I get home. Thanks for reviewing.

Daggolin (approved):

SomaZ (approved):

PR #1253, opened 2025-02-21
Fix spawn item error va eval

SomaZ:
There's just two arguments but three are expected. so this will hard crash.


SomaZ:
Crash can be triggered trying to load this map in SP:
https://jkhub.org/files/file/213-blueice-twilight/

Razish (approved):

taysta (approved):

Daggolin (approved):

PR #1255, opened 2025-03-01
[SP] Force power Timing Fix

BE-Arbiter:
Fix Force power activating to many time per second at high fps.

### DIFF
--- a/code/game/g_public.h
+++ b/code/game/g_public.h
@@ -108,6 +108,7 @@ struct gentity_s {
 
 	vec3_t		absmin, absmax;		// derived from mins/maxs and origin + rotation
 
+	int			forcePowerTime;		//Delay between force power activation
 	// currentOrigin will be used for all collision detection and world linking.
 	// it will not necessarily be the same as the trajectory evaluation for the current
 	// time, because each entity must be moved one at a time after time is advanced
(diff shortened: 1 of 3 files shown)


ensiform:
This adds new fields to saves breaking compat, however.

BE-Arbiter:
Having a better alternative ?

ensiform:
Don't put in playerstate

BE-Arbiter:
Where would you put it ? Isn't ent->s also save breaking? In a logic standpoint it's where all force properties are stored and its not like a table of size max_entities would be an elegant solution...

ensiform:
> Where would you put it ? Isn't ent->s also save breaking? In a logic standpoint it's where all force properties are stored and its not like a table of size max_entities would be an elegant solution...

Any reason why it has to be in the entity state or player state at all, not just part of the client struct?

BE-Arbiter:
it's more clean to my psychorigid mind. Will make your proposed adaptations

BE-Arbiter:
Is it better now ? :)

ensiform:
Well it will not compile, you put it in entity structure and reference it in client structure.

BE-Arbiter:
There.

ensiform:
Most likely not going to merge this.

Drain, lightning and rage all have denouncers in ojk multiplayer that are correctly based on the sv_fps of 20 instead. Magic number +33 is a bit odd.

BE-Arbiter:
And instead of telling me you don't like this solution you play "guess my thought" for two days ?

Have fun.

ensiform:
Took me a while to track it down. 

https://github.com/JACoders/OpenJK/blob/master/codemp%2Fgame%2Fw_force.c#L4133

https://github.com/JACoders/OpenJK/blob/master/codemp%2Fgame%2Fg_local.h#L794-L798

PR #1262, opened 2025-04-29
Fix portable build on linux/macos

aufau:
Portable build makes current directory fs_homepath
Apparently it has never been implemented in sys_unix.cpp, only in sys_win32.cpp


taysta:
Tested portable build on macOS with this commit cherry picked, works as expected

Razish (approved):

SomaZ (approved):

Daggolin (approved):

PR #1281, opened 2025-08-25
Reorder init to match MP startup

### LINKED ISSUE #897: LucasArts intro video has distorted audio
(Arch) Linux 64-bit, OpenJK (JK2) built from Git master on 19 November.

Single player.

The JK2 SP intro video with the LucasArts logo exhibits slightly distorted audio. I can still make out the sound but it seems to skip audio in intervals of about ~75 ms.

I'm using PulseAudio, not sure if it's related. If there is any other information you need that may be useful in debugging such an issue, I'd be happy to provide it. It may be worth mentioning that the LucasArts intro video in JA _does_ sound fluent and is not distorted.

It's not really a major issue or one that bothers me in particular, but I thought I'd report it anyway.

The sound to play fluently.

TwelveEyes:
Reorders SP startup init calls to match the order of the MP binary. Fixes the audio distortion during the LucasArts intro video.

Fixes #897

### DIFF
--- a/code/qcommon/common.cpp
+++ b/code/qcommon/common.cpp
@@ -1094,11 +1094,11 @@ void Com_Init( char *commandLine ) {
 		//Swap_Init ();
 		Cbuf_Init ();
 
-		Com_InitZoneMemoryVars();
-		Cmd_Init ();
-
 		// override anything from the config files with command line args
 		Com_StartupVariable( NULL );
+
+		Com_InitZoneMemoryVars();
+		Cmd_Init ();
 
 		// done early so bind command exists
 		CL_InitKeyCommands();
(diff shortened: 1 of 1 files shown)


code/qcommon/common.cpp
```diff
@@ -1196,6 +1196,7 @@ void Com_Init( char *commandLine ) {
 				Cbuf_AddText ("cinematic openinglogos\n");
 			}
 		}
+		CL_StartHunkUsers();
```
TwelveEyes, line 1199: I'm unsure this call is necessary. I only added it so SP and MP match 1:1.

ensiform: That is the only part of what makes the fix happen, the moving of the other two functions has no effect on the sound code.

ensiform:
In a perfect world, would just remove SP engine and make SP features work in "MP" sans a real server connectivity.

PR #1283, opened 2025-08-28
Fix missing defaults for `fx_rain` and broken weather zone creation

### LINKED ISSUE #1104: Jedi Outcast Weather Problems
Not a huge deal but thought I'd bring it up.

For OpenJO, weather effects seem to not work very consistently with wind missing and rain/snow sometimes passing through geometry. This is apparent in yavin_swamp and the demo map.

Saving then loading also seems to cause weather to clip through all geometry.

TwelveEyes:
`lightrain` and `wind` added to `fx_rain`  since it was missing defaults if it was created without any spawn flags (since the JO weather system doesn't use spawn flags). JO's rain system always has random wind, unlike JA's rain system. Enabling `wind` is close enough for this purpose.

JO code:
https://github.com/grayj/Jedi-Outcast/blob/master/code/game/g_fx.cpp#L276-L287
https://github.com/grayj/Jedi-Outcast/blob/master/code/renderer/tr_worldeffects.cpp#L1986-L2003

Auto generated weather zones were also not being created properly beyond first map load because of bad operation placement, resulting in weather effects clipping through geometry. Currently, if a cache file is found and `mWeatherZones` is empty (not all maps have pre-placed weather zones, for example JO's yavin4) then `mWeatherZones` will never be filled, and the point data from the cached file is useless. Weather zones should always be generated if `mWeatherZones` is empty (whether or not a cache file is found is irrelevant), as the cache file only contains point data, not weather zones themselves.

Fixes #1104

### DIFF
--- a/code/game/g_fx.cpp
+++ b/code/game/g_fx.cpp
@@ -563,6 +563,12 @@ void SP_CreateRain( gentity_t *ent )
 		G_EffectIndex( "world/acid_fizz" );
 		G_FindConfigstringIndex("acidrain", CS_WORLD_FX, MAX_WORLD_FX, qtrue);
 	}
+	else
+	{
+		G_FindConfigstringIndex("lightrain", CS_WORLD_FX, MAX_WORLD_FX, qtrue);
+		G_FindConfigstringIndex("wind", CS_WORLD_FX, MAX_WORLD_FX, qtrue);
+		return;
+	}
 
 
 	// OUTSIDE SHAKE
--- a/code/game/g_fx.cpp
+++ b/code/game/g_fx.cpp
@@ -567,7 +567,6 @@ void SP_CreateRain( gentity_t *ent )
 	{
 		G_FindConfigstringIndex("lightrain", CS_WORLD_FX, MAX_WORLD_FX, qtrue);
 		G_FindConfigstringIndex("wind", CS_WORLD_FX, MAX_WORLD_FX, qtrue);
-		return;
 	}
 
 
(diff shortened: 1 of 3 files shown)


Razish (dismissed):

code/game/g_fx.cpp
```diff
@@ -563,6 +563,12 @@ void SP_CreateRain( gentity_t *ent )
 		G_EffectIndex( "world/acid_fizz" );
 		G_FindConfigstringIndex("acidrain", CS_WORLD_FX, MAX_WORLD_FX, qtrue);
 	}
+	else
```
SomaZ, line 566: This changes behavior if prior flags aren't set. I'd prefere a if (ent->spawnflags == 0) *add_defaults; at the start of the function.

SomaZ (changes requested):

code/game/g_fx.cpp
TwelveEyes: I organized the conditions to match the convention in `SP_CreateSnow` (https://github.com/JACoders/OpenJK/blob/master/code/game/g_fx.cpp#L301-L305). It can be changed of course, but it would be odd to me to have the conditions organized two different ways.

SomaZ: You didn't because you added a return that will prevent the evaluation of the remaining spawnflags. So removing the return should be fine

TwelveEyes: Oops, removed that return.

SomaZ (approved):

Razish (approved):

Issue #1282, opened 2025-08-27
Jedi Outcast (JO) lightsaber segfault on linux Debian


NilsEckerle:
# Report

Segfault after selecting lightsaber after using cheats in first level.

- OS:On Debian 13.
- DE/WM: i3wm
- GPU: AMD Radeon RX 7600 XT
- CPU: AMD Ryzen 5 7600X 6-Core Processor
- gcc: gcc (Debian 14.2.0-19) 14.2.0
- commit hash: `2f6abb4f` (current one on main branch)

## How to reproduce the error

After building the game and installing it the game works fine, but when im in
the starting level and cheat all weapons `helpusobi 1`, `give all`, the game
crashes when selecting the lightsaber.

when the light saber is obtained in the level BESPIN_STREETS which you can get
with the cheat `map BESPIN_STREETS` the lightsaber works without crashes.

## Step by step

1. Clone the repository `git clone https://github.com/JACoders/OpenJK.git ~/OpenJK/OpenJK`
2. Get the Jedy Outcast GameFile into `~/OpenJK/GameFile`
3. Building the project:
    - change CMakeList.txt option `BuildJK2SPEngine`, `BuildJK2SPGame` and `BuildJK2SPRdVanilla` to `ON`
    - `cd ~/OpenJK/OpenJK && mkdir build && cd build`
    - `cmake -DCMAKE_INSTALL_PREFIX=$HOME/OpenJK/GameData ..`
    - `make && sudo make install`
4. Move files
    - `cd ~/OpenJK/GameFile`
    - Delete unnecessary files `rm *.exe *.dll`
    - Move files `sudo mv JediOutcast/* .`
5. Run the program `./openjo_sp.x86_64`
6. Setup segfault ingame
    - New > Jedi > Begin Game > Wait untill ingame
    - Open console \[\`/~ or ^ button\]
    - Enable cheats `helpusobi 1`
    - Give all weapons `give all`
    - Press `1` to switch to light saber **(Segfault here)**

## Expected Behavior
Lightsaber should be selectable without crashing.

## Actual Behavior
Game crashes with segfault when pressing `1` to select lightsaber.

# NOTE
When entering level BESPIN_STREETS with `map BESPIN_STREETS` the lightsaber works.

<img width="1223" height="1403" alt="Image" src="https://github.com/user-attachments/assets/45078ee2-bcf4-49c9-b03c-000ee59bdfd4" />

[...]

PR #1284, opened 2025-08-28
feat(build): add BuildReleaseVersion option

NilsEckerle:
Add a build release version flag to CMakeLists.txt to make clear this is a option to set.

In the issue #1282 i ran into an error while building via cmake. This would have fixed it.


ensiform:
Completely unnecessary, pass argument on command line.

NilsEckerle:
I still think it is important to show that `-DCMAKE_BUILD_TYPE=Release` is something that can be set. I didn't find it so I had to open the issue. Maybe adding this to the Compilation-guide is a solution. @ensiform

PR #1294, opened 2025-11-18
[MP] Fix memory leak when loading sub BSPs

Alereon:
In `CM_LoadMap_Actual`, a temporary buffer is allocated via `Z_Malloc` to hold the raw BSP file data. For the main map, this buffer is cached in `gpvCachedMapDiskImage`. However, for sub BSPs, the buffer is allocated but never freed, leaking memory on every sub BSP load.

This fix frees the temporary buffer for sub BSPs after parsing while preserving the main map caching.


Daggolin (approved):
Looks good.

As this fix affects servers and clients when using misc_bsp it's one that should get pulled downstream into client forks as soon as it's merged here. Tagging @taysta for TaystJK.

AlexCS1337 (approved):

ensiform (approved):

PR #1310, opened 2025-12-01
[SP] Fix animation event sounds not playing for NPC variants

### LINKED ISSUE #1287: JA SP -  Stormtrooper Death “Clank” Sounds Don’t Play for NPC Variants
I think this is a vanilla bug, but I noticed in singleplayer JA, Stormtrooper variants don't play their custom death sounds (bodyfall_trooper1-4.mp3) defined in the model's "animevents.cfg" file. 

This affects the NPCs:
- stormtrooper2
- stofficer
- stofficeralt
- stcommander
- rockettrooper

They all use the same "playermodel", but the sounds only seem to trigger for the base "stormtrooper" NPC.

dpiers:
## Summary

Animation event sounds (like stormtrooper death clanks) weren't playing for NPC variants such as `stormtrooper2`, `stofficer`, etc.

## Cause

Model-specific animation events are loaded from `models/players/<model>/animevents.cfg` and tagged with `modelOnly = hstring("<model>").handle()` to restrict them to that model.

At runtime, `CG_PlayerAnimEvents` compared this against `NPC_type`, but NPC variants like "stormtrooper2" use the "stormtrooper" playermodel while having a different `NPC_type`. So the event's `modelOnly` (from "stormtrooper") didn't match `myModel.handle()` (from "stormtrooper2"), and the sound was skipped.

## Fix

Use `level.knownAnimFileSets[animFileIndex].filename` - this is the model name that was used when loading the animation events, so it will always match.

Fixes #1287

### DIFF
--- a/code/cgame/cg_players.cpp
+++ b/code/cgame/cg_players.cpp
@@ -1220,7 +1220,9 @@ static void CG_PlayerAnimEvents( int animFileIndex, qboolean torso, int oldFrame
 		}
 	}
 
-	hstring myModel = g_entities[entNum].NPC_type;		//apparently NPC_type is always the same as the model name???
+	// Use the anim file set's model name instead of NPC_type, since NPC variants
+	// (e.g., "stormtrooper2") share the same model and anim events as the base NPC
+	hstring myModel = level.knownAnimFileSets[animFileIndex].filename;
 
 	// Check for anim event
 	for ( i=0; i < MAX_ANIM_EVENTS; ++i )


ensiform:
As of right now we have no intention of accepting any of these AI generated PRs.

dpiers:
Why? 


I can understand declining the pro-active memory safety fixes but this one and #1309 are clear and well documented bug fixes with open issues.

ensiform:
1309 produces incompatible saves just by virtue of your change. And a majority of alleged memory safety ones are false positives/unnecessary/incorrect fixes even if they produce code that compiles and runs, such as the msg netcode change in MP using SP as an example which Claude apparently didn't fully understand why the codebases differ and what potential issues it leads to including your change which will truncate network messages on server or client side and this adds a network incompatibilities, the existing code was safe as it is.

Others may chime in as well but the jist is that we do not feel comfortable with the drive-by mass PRs generated by AI and are leaning towards blocking all AI generated commits anyway. In this instance it gives off vibes that you may attempt to sneak malicious code in see: xz takeover situation for example. We will still probably look further if they can be accepted but it may take time.

dpiers:
Sorry - I didn't mean to PR spam. I am still learning to use/work with Claude Code and thought helping improve this codebase would be good practice, and it started automatically sending PRs before I realized it.

I actually started out having Claude look at my own Jedi Academy/Outcast repos, which were among the first on GitHub in April 2013 after the original source was made available on SourceForge, and then found this project.

I was a game engine programmer on Star Wars: Kinect and feel like helping fix some ~22 year old bugs here is the least I can do to repay my debt to the SW gaming community.

FWIW: The PRs that are still opened were actively investigated, reproduced and had the fixes tested.

DisrespectDwardo:
Delete your account, this provides no value.

PR #1311, opened 2025-12-02
[MP] Fix ambient sounds not playing after snd_restart

### LINKED ISSUE #1063: Ambient sound does not work after map change or audio reinitialization, but works if vid_restart
Windows 10 64-bit and macOS Catalina 10.15.7

Multiplayer

After a map change or after some time in-game, audio from elevators with soundSet platform or doors with soundSet impDoor1 do not play any audio. Music continues to work, footsteps continue to work, sabers and weapons continue to work, voice continues to work.

If I switch EAX on/off (no matter if it was working before or not), it ceases to work until a vid_restart is performed. It happens on both my mac and windows machine. I just downloaded the latest December 2020 build because I thought maybe my compile was bad from a while ago, but it still happens.

soundSets working correctly throughout the game.

dpiers:
Fixes #1063

## Root Cause

`AS_Init()` dynamically allocates `pMap`. When `snd_restart` runs:

1. `S_Shutdown()` → `AS_Free()` sets `aSets = NULL` but leaves `pMap` intact
2. `S_Init()` → `AS_Init()` sees `aSets == NULL`, allocates a **new empty `pMap`**, discarding all soundset names cgame had registered
3. `AS_ParseSets()` finds an empty `pMap` and loads nothing

The SP codebase already uses a static singleton for `pMap` (has since original Raven source), avoiding this issue.

## Affected Sounds

All 166 ambient sound sets in `sound/sound.txt`: bmodel sounds (doors, platforms, elevators), general environmental ambience, and local positional sounds.

## Fix

- Use a static singleton for `pMap` in MP (matching SP implementation)
- Call `AS_ParseSets()` in `CL_Snd_Restart_f()` to reload ambient sets after restart

## Testing

1. `devmap mp/ffa5`
2. Verify door/ambient sounds work
3. `snd_restart`
4. Confirm sounds still play (previously went silent)

### DIFF
--- a/codemp/client/cl_main.cpp
+++ b/codemp/client/cl_main.cpp
@@ -1267,6 +1267,9 @@ void CL_Snd_Restart_f( void ) {
 
 	extern void S_RestartMusic( void );
 	S_RestartMusic();
+
+	extern void AS_ParseSets( void );
+	AS_ParseSets();
 }
 
 
--- a/codemp/client/snd_ambient.cpp
+++ b/codemp/client/snd_ambient.cpp
@@ -758,13 +758,21 @@ Loads the ambient sound sets and prepares to play them when needed
 -------------------------
 */
 
+static namePrecache_m *TheNamePrecache( void )
+{
+	// Use a static singleton so pMap persists across sound restarts.
+	// This fixes ambient sounds not playing after snd_restart.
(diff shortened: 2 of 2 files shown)


DisrespectDwardo:
Not a real issue, delete your account.

PR #1335, opened 2026-06-03
Improve new macOS icons

Circa-III:
SomaZ's icons from #1272 had a white background by default, so I matched them better to the Windows icons. The OpenJO icon also appeared incorrectly in Finder at smaller sizes because I believe some icon resolutions were missing from the .icns file. That is fixed now too. (ignore the build_ded.yml file. I did not mean to include that in this.)

openjk-icons-new


SomaZ (approved):

taysta (approved):
Tested locally on mac, fixes the previously broken icons.

ensiform (approved):

PR #1340, opened 2026-06-30
Patched performance bug in quickload behaviour

tayloroxelgren:
This fixes what appears to be a quickload performance bug in Jedi Outcast single player.

During JK2 SP sound precache, the loading screen text was updated for nearly every sound entry :

```cpp
if (i&31) {
    CG_LoadingString( soundName );
}
```

I believe this was just a long standing bug as the Jedi Academy code has similar code but it is written like so.

```cpp
if (!(i&7)) {
      CG_LoadingString( soundName );
}
```

In my local testing, it resulted in a **20-25%** reduction in quickload times and **2-5%** for cold loads.

### DIFF
--- a/codeJK2/cgame/cg_main.cpp
+++ b/codeJK2/cgame/cg_main.cpp
@@ -752,7 +752,7 @@ static void CG_RegisterSounds( void ) {
 		if ( soundName[0] == '*' ) {
 			continue;	// custom sound
 		}
-		if (i&31) {
+		if ((i & 31) == 0 ) {
 			CG_LoadingString( soundName );
 		}
 		cgs.sound_precache[i] = cgi_S_RegisterSound( soundName );


ensiform:
There is no real measurable performance gain here that I can tell. Flipping the logic from 31 out of 32 sounds to only printing every 32 sounds doesn't save very many draw calls. It does not result in fewer registrations being made which is what really takes the time on load.

tayloroxelgren:
I just reran the quickload test with and without this change.

Each case was run 10 times on `kejim_post`.

Without change:
```
TOTAL LOAD TIME avg  : 483.2ms
CG_RegisterSounds total avg : 132.0ms
```

With the change:
```
TOTAL LOAD TIME avg : 359ms
CG_RegisterSounds total avg : 8ms
```

I agree it doesn't reduce the number of registrations. The difference comes from reducing `CG_LoadingString()` calls during quickload when the sound registration has already been cached. So just calling `CG_LoadingString()` and updating the screen is inflating the quickload time.

PR #1341, opened 2026-07-10
Add local split-screen multiplayer prototype

joshabb:


ensiform:
Great proof of concept!

One thing to note, since it utilizes mod code changes you will see much fewer usage as you won't be able to play custom mods that are out of support or don't include it. Did you look at zturtleman/spearmint for inspiration for this?

PR #1344, opened 2026-07-12
Fix bundle SDL2 for macOS releases

Circa-III:
Build SDL2 2.32.10 from source for macOS release builds and bundle it inside each generated application bundle.

- Build a pinned SDL2 release during CI
- Link OpenJK against the source-built SDL2
- Bundle libSDL2-2.0.0.dylib into each .app
- Preserve internal libraries for portable builds while excluding the Windows-only bundled SDL2 libraries
- Verify the bundled SDL runtime during CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -227,7 +227,7 @@ jobs:
           if [ "${{ matrix.portable }}" == "Portable" ]; then
             OPTIONS+=" -DUseInternalLibs=ON -DBuildPortableVersion=ON"
           else
-            OPTIONS+=" -DUseInternalLibs=ON -DBuildPortableVersion=OFF"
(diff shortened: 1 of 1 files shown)


Circa-III:
Ditched homebrew altogether for macos SDL2, uses a pinned SDL2 build and bundles that. Tested on macOS 27 Dev Beta 4 on a MacBook Pro (M3 Pro). I do not have an older intel Mac to test with.
Screenshot 2026-07-29 at 2 16 12 PM
Also tested other artifacts on ubuntu 64bit and windows 11 just in case I broke something on those OSes but they launch fine.

Circa-III:
The whole build system is a pain in the ass but I got it. After discussing on discord, some of the devs prefer we pull SDL2 from the release DMG itself rather than build our own every time, so this final version does just that. Tested OpenJK and OpenJO, all should be good now.

PR #1345, opened 2026-07-13
[CG] Fix cg_fovViewModel when cg_fovAspectAdjust

SomaZ:
The scaling factor should always be the same and based on the difference of the set fov and the wanted fov. All fov effects can work properly now and arent overwritten and it works with aspect correction.

### DIFF
--- a/code/cgame/cg_weapons.cpp
+++ b/code/cgame/cg_weapons.cpp
@@ -1126,9 +1126,10 @@ void CG_AddViewWeapon( playerState_t *ps )
 	AnglesToAxis( angles, hand.axis );
 
 
-	if ( cg_fovViewmodel.integer ) {
-		float fracDistFOV = tanf( cg.refdef.fov_x * ( M_PI/180 ) * 0.5f );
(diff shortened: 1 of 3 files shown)


ensiform:
Doesn't this break cases where cg_fov is overridden

SomaZ:
> Doesn't this break cases where cg_fov is overridden

This is exactly what it actually fixes. Currently, cases with overwritten fov will still force the exact defined weapon fov. This is not desirable though as you want the effect to apply to the weapon as well.

SomaZ:
Running cg_fov 100 and cg_fovViewModel 0, using force speed:
shot2026-07-15_00-14-50
shot2026-07-15_00-14-39

running cg_fov 100 and cg_fovViewModel 101 with fixed code:
shot2026-07-15_00-15-10
shot2026-07-15_00-15-16

running cg_fov 100 and cg_fovViewModel 101 with broken code:
shot2026-07-15_00-26-17
shot2026-07-15_00-26-30

now broken code with disabled cg_fovViewModelAdjust to get the model at least in sight:
shot2026-07-15_00-26-48
shot2026-07-15_00-27-02
shot2026-07-15_00-33-01
shot2026-07-15_00-26-30

PR #1352, opened 2026-09-15
Fix/hl2 weapon sway

joshuahelm088:


joshuahelm088:
Closed: opened against upstream by mistake. This change belongs in the personal fork only.

